PR #6, opened 2016-03-09
Added O3PipeView trace generator.

s-okai:


src/main/scala/dpath.scala
```diff
@@ -1177,9 +1215,7 @@ class DatPath() extends Module with BOOMCoreParameters
 
    // Time Stamp Counter & Retired Instruction Counter
    // (only used for printf and vcd dumps - the actual counters are in the CSRFile)
-   val tsc_reg = Reg(init = UInt(0, xLen))
-   val irt_reg = Reg(init = UInt(0, xLen))
-   tsc_reg := tsc_reg + UInt(1)
+   tsc_reg := tsc_reg + Mux(Bool(O3PIPEVIEW_PRINTF), O3_CYCLE_TIME, UInt(1))
```
ccelio: What's O3_CYCLE_TIME, and why is it 500?

ccelio:
Sorry, I've removed datpath.scala and moved it into core.scala. Can you submit a new pull request that's rebased off the latest master? It should be fairly straightforward to copy everything over, even just doing it manually. It's a little hard to parse the pull request since it includes things that have changed on master.

If not, I can do it myself, but it'll be a few days before I have the free cycles.

Otherwise, it looks great and I'm excited to try it out!

src/main/scala/dpath.scala
s-okai: In the gem5 pipeline viewer, you can set a number of ticks per clock such that it matches the CYCLE_TIME used for the gem5 simulator (see http://www.m5sim.org/Visualization). That constant is just used so that the timestamps match the gem5 convention when generating traces for the pipeline viewer.

s-okai:
Hi,
I have some time, so I can make the updates by hand.

ccelio:
Thanks!

Can you also add to the README a new section/paragraph on how to use the pipeline visualizer?  What tools are required, where to get them, etc.  This is the kind of feature you don't want to hide.

s-okai:
No problem

s-okai:
Took care of moving things over to core.scala, and the README has been updated with instructions on how to use the visualizer with BOOM. Let me know if you need anything else from me.

ccelio:
A couple of comments:

First, 

Did you look at the generated output? It looks to be fairly nonsensical, often the order of stages is borked:

```
dpfni...cr
```

It looks close to being right, but not quite there. :/

Second, 

Did you notice a significant increase in the compile time of BOOM when adding this in (particularly the gcc compile step)?  I'm trying this with `O3PIPEVIEW_PRINTF=false`, and I'm seeing the Chisel node count has risen to `[info] [109.054] compiling 253653 nodes`, which is an 8% increase from before.

The full "make run" test has gone from <6 minutes to >8 minutes (again, with the pipeview_printf turned off).  =(

I'll look into the broken behavior, and I'll merge this in once that's fixed. However, I'm concerned about the compile/performance hit when the feature isn't being used, and I'd really like to refactor this to be a "pay only when you use it" feature.  Off the top of my head, the typical trick is to leverage Scala if/else conditions to not generate Chisel Nodes.  I suspect a big part of the problem may be putting the StageEvents in the MicroOp bundle, which means that EVERY MicroOp in the machine may be paying this penalty. The solution may be to store StageEvents exclusively in the ROB, separate from the MicroOp bundle. This will require some more pondering on my part, since it seems like it's pretty much close to this design already.

ccelio:
I think the incorrect behavior is that Fetch/Decode/Rename/Dispatch are all being given the same cycle time.

s-okai:
Is there any way to ifdef out the timestamps from the MicroOp bundle? Its easy enough to do everywhere where the time-stamped are used, but I'm going to guess we'll see most of the performance hit as long as its allocated.

I also looked into the same timestamp being logged for several stages. Not quite sure what would cause this since  I assume there are registers between where I log the timestamps in each stage. Might be a misunderstanding on my part for how Chisel works, but it's not immediately obvious to me why this happens.

Either way I'll take a look.

src/main/scala/core.scala
```diff
@@ -411,6 +437,8 @@ class BOOMCore(implicit p: Parameters) extends BoomModule()(p)
                                 flow=p(EnableFetchBufferFlowThrough),
                                 _reset=(fetchbuffer_kill || reset.toBool)))
 
+   fseq_reg := fseq_reg + Mux(FetchBuffer.io.enq.valid, UInt(FETCH_WIDTH), UInt(0))
```
ccelio, line 440: Is the fetch sequence number used for anything, or is it just a nice, pretty piece of information in the printout? I'm looking at the output logs, and this is appears to be incrementing a bit too much.  For starters, perhaps something like a PopCount() of the "mask" bits should be used instead?

ccelio:
Hmmmm, I think I better understand the timing issue now. 

The `fetch_tsc` is being set too late - you're marking the fetch_bundle as it exits the icache, but we should really mark it for when we initiate our _request_ to the icache for that instruction. The easy solution is to just set `fetch_tsc` to `tsc-UInt(2)`, since the Fetch Pipeline goes "send PC to icache -> access icache data RAMs -> return fetch_bundle".   Although there's one rub - it won't properly handle an icache miss where we make a request to the icache and the icache must go and fetch the data (or there's an iTLB miss).  That's harder to solve though, since the obvious solution involves modifying the icache code, which is a part of Rocket and not BOOM.

The second issue is when the fetch_bundle comes back from the icache, it is put into the FetchBuffer queue.  However, that's a flow-through queue, so if it's empty, the data bypasses the queue and goes straight to the Decode stage within the same cycle.  Also, Rename executes in parallel with the Decode stage, and the Dispatch stage is actually squeezed into a single stage with Decode/Rename/Dispatch.  Therefore, it's possible for Fetch+Decode+Rename+Dispatch to all have the same time stamp count!  I'm disappointed to see that gem5 gets confused by this, and instead lists the stages as occurring in some random order on different cycles, instead of dropping stages and only listing out the first to occur on a given cycle.

I'm not sure what the solution is.  We may have to (horror of horrors) make a change to the gem5 script to handle this scenario properly. We can either make a pull request to gem5, or we might be able to fork their code and include the scripts in the BOOM repository.  I've only made a cursory glance at the code, but I think it's a BSD license, so maybe we can modify it and add it to the BOOM repo.

src/main/scala/core.scala
s-okai, line 440: See line 1509. It is used for the sequence number for the trace.

That being said, I'm a little confused how at PopCount() on the mask bits would be better (or even correct)? Isn't the entire width valid or not valid, in which case we should either increment by FETCH_WIDTH or not at all?

ccelio, line 440: Right, but is that of functional importance to the o3-pipetrace script, or could it be any number we wish? I ask because the fseq_reg is currently inaccurate as it is, but if it's not a functional thing, it's okay to leave it as is.

s-okai:
We could look into fixing this in gem5 itself. Its probably an easy fix and probably worth doing. I don't completely like forking it off and having an independent version only because:
1. This should actually be fixed in gem5.
2. Probably better if we don't have to maintain the visualization generator.

ccelio:
I'm looking at gem5/src/cpu/o3/dyn_inst_impl.hh, and it looks like we're suppose to print out "0" for stages that "don't exist", if you will. Pipeview isn't smart enough to handle stages that are the same cycle, or handle stages that don't print out to the file, but if you write "0" instead it seems to work as expected!

This also means that we could print out misspeculated instructions that are killed before commit, but that admittedly makes this a tad more challenging (what are we suppose to do? have a for loop through the entire ROB that detects instructions killed on any given cycle and prints them out? yuck). 

You know, there may be a completely different way to implement this in BOOM.  What if, instead of storing the StageEvents in each micro-op in the ROB, and printing all stages out together when the instruction is committed (which means we're storing a lot of stuff, passing a lot of signals around the processor, and it means we miss un-committed instructions! yuck!), what if we printf each micro-op as it passes each stage?  

Instead of one big printf in commit, we have ~7 printfs, each one in a different stage of the processor. The only thing we pass around is the fseq number.  

Now, the only problem is o3-pipeview.py is expecting to receive each instruction in sequence, with all stages printed out together.  One solution is we write our own middle-ware script: we print out the fseq number in front of every printf, and then re-assemble each instruction using the fseq tags until it's in the format that o3-pipeview.py wants. I'm not particularly a fan of adding more scripts and making the user (or Makefile) handle piping stuff around, but I'm not sure of how we're suppose to handle misspeculated instructions in any other way.

s-okai:
Does the visualizer display misspeculated instructions? I know I have definitely seen it skip sequence numbers (in fact the screenshot on the site shows it http://www.m5sim.org/Visualization), and I assume that means that it flushed the instructions in between. Currently I only print instructions that are committed, and I can't think of why'd we'd want to print misspeculated instructions if the pipeline viewer throws them away anyways. If someone wanted to look at the trace by hand it might be useful, but I don't think anyone would considering its very hard to read through it. As far as I can figure, logging misspeculated instructions would also only increase the overhead further (unless of course we go with a middle-ware script).

Also from what I understand, you're saying if there are multiple stages with the same timestamp, if we all but one of them, everything is displayed as expected? That is a change I could put in, and requires very little code.

ccelio:
The pipeview script has an option to only print committed instructions, which implies it by default prints all fetched instructions. 

```
'only_committed':0,   # Set if only committed instructions are printed. 
```

It also has code like:

``` python
# Find out the time of the last event - it may not                                 
# be 'retire' if the instruction is not comlpeted.                                 
last_event_time = max(inst['fetch'], inst['decode'],inst['rename'],                
                  inst['dispatch'],inst['issue'], inst['complete'], inst['retire'])
....
# This will visually distinguish completed and abandoned intructions.  
if inst['retire'] == 0: dot = '=' # abandoned instruction              
else:                   dot = '.' # completed instruction              
```

So it's clearly intended to show misspeculated instructions.  

> Also from what I understand, you're saying if there are multiple stages with the same timestamp, if we all but one of them, everything is displayed as expected? That is a change I could put in, and requires very little code.

I have it writing "zero" into the rename and dispatch tsc's, and I currently set fetch_tsc to be -UInt(2).  

I'm hoping to merge in your changes (with my edits) tonight more or less as is, but I do believe a re-think on the implementation may be required. There's a few tricks we may be able to play - for example, it seems that o3-pipeview is able to handle instructions written to it out-of-order, and it uses the fseq to re-order instructions.

ccelio:
I also believe the store completion logic is incorrect - it's suppose to mark when the store is sent to memory (which happens after the store has been committed), and not when the store tells the ROB it _can_ be committed. 

There's extra logic flag to o3-pipeview that enables the store-completion print-outs, which defaults to false.

src/main/scala/rob.scala
```diff
@@ -370,6 +393,8 @@ class Rob(width: Int
       when (io.lsu_clr_bsy_valid && MatchBank(GetBankIdx(io.lsu_clr_bsy_rob_idx)))
       {
          rob_bsy(GetRowIdx(io.lsu_clr_bsy_rob_idx)) := Bool(false)
+         rob_uop(GetRowIdx(io.lsu_clr_bsy_rob_idx)).debug_events_tsc.memory_tsc := io.tsc   // Record store and complete timestamps
```
ccelio, line 396: This appears to be marking both loads and stores with the "memory_tsc"? That doesn't seem correct. I also believe the "store_tsc" is suppose to track when the store is issued to the data cache, which is well after it is committed.

src/main/scala/core.scala
s-okai, line 440: Yes, it is important to the script, as it is the sequence number that the script uses to order the instructions. I'm still not quite sure I understand why it is inaccurate.

ccelio, line 440: For starters, it's not always going to fetch the full FETCH_WIDTH of instructions (like if it runs off the end of the cache line, or it predicts a branch is taken in the middle of the packet). 

You'd have to look at it in a waveform viewer to see other scenarios where `FetchBuffer.io.enq.valid` is true, but otherwise no progress is made. I think at least for that case, it needs to also check `FetchBuffer.io.enq.ready`, since the FetchBuffer may be full and not accepting more packets.

src/main/scala/rob.scala
s-okai, line 396: Looking at the traces generated from the gem5 sims, you're right about this. May need to rethink this since we cannot simply print on a commit. Easiest way is probably to ignore commits that are stores (easy enough to find by checking the fields in the MicroOp) and print them in the LSU when sent to memory instead. The only issue with this is that the ROB needs to notify the LSU when it commits a store. Since stores are sent to memory in commit order, we can probably just push the commit timestamps to a queue and pop one out each time a store is sent to memory.

s-okai:
Being able to print out of order is probably useful for the case where we print flushed instructions, since we can just blindly print an instruction as soon as its flushed. I'm not sure that it is useful outside of that unless we go with the middleware option since we still have to wait for a commit timestamp before we can print (in which case we still print in order) if the instruction does not get flushed.

src/main/scala/rob.scala
ccelio, line 396: The ROB already has to tell the LSU to commit a store; otherwise the store wouldn't know when it's allowed to be sent to memory.  There's a commit bit in the store queue, so each store knows when it's been committed.  If we're eschewing adding a post-processing script, this sounds like a reasonable solution.

ccelio:
I believe the o3pipeview script can handle OoO instructions, by way of the fseq number.  The requirement for a middleware script comes in if we print out the instructions piece-meal over many cycles and interleave instructions.  The advantage of this is each MicroOp only needs to track its fseq number, so it's lower overhead, and it easily provides us the misspeculated instructions.  

My current measurements shows that this pull request adds 10% to the Chisel Node Count and the compilation time, so I'm fairly sensitive to using techniques that lower the state per MicroOp.

src/main/scala/core.scala
s-okai, line 440: Ah ok I see now. So I assume a better choice would be:

```
fseq_reg := fseq_reg + Mux(FetchBuffer.io.enq.valid && FetchBuffer.io.enq.ready, PopCount(FetchBuffer.io.enq.bits.mask), UInt(0))
```

s-okai:
A few easy fixes:
1. Remove the StageEvents() bundle from FetchBundle, and only leave the fetch timestamp. We definitely do not need the all the timestamps in the frontend.
2. Leverage Chisel if's to remove some logic when we disable the trace. It seems they are at the very least allowed in bundle definitions (things seem to compile and run with them) so this should remove all the overhead when the trace is turned off.

Let me know what you think.

ccelio:
I'm struggling to make Scala conditionals work.  I've tried if() new StageEvents() else null, but null is causing errors deep in the Chisel compiler; it doesn't like to make comparisons across I/O bundles. I've tried Option, which seems the most promising, but that's causing missing Wire() declarations when using uop.debug_events.get.fetch_tsc on the left-hand side of a Chisel assignment. 

I have a few more ideas to pursue here, but I'm currently thinking of completely removing StageEvents from MicroOp, and instead having essentially a separate ROB of just StageEvents.

``` scala
val stage_events = Mem(NUM_ROB_ENTRIES, new DebugStageEvents)
```

Each uop after decode has an rob_idx, and on dispatch it writes in the fetch/decode/dispatch_tsc's. 

It will become a highly ported Mem though (dispatch_width+commit_width+issue_width+wb_width, although I could just move the StageEvents into the actual ROB too, if I wanted a more optimal design), this technique falls apart for handling store-completions, and it doesn't really help handle misspeculated instructions.

s-okai:
One last thing to try is lazy val:

```
lazy val debug_events_tsc = new StageEvents    // stage event timestamps
```

when defining MicroOp.

I tried adding a dummy field using this method, and it seems to reduce the node count if the field is not ever used. You'd have to put if's (around anything that accesses MicroOp.debug_events_tsc to see any benefits, but might be worth a try.

s-okai:
Down to 251216 nodes using lazy val and using if's

ccelio:
Hmmm, lazy was a good suggestion, but it seemed to still give me errors if O3PIPEVIEW_PRINTF was set to false. 

I've found a solution though: it turns out that a recent Chisel update allegedly adds Option support within Bundles, and I may have been interfacing with the Some field incorrectly earlier anyways.

I'll finish my merge of the pull request, but going forward, I think a re-think may be required to support store-completion, misspeculated instructions, and icache misses.

ccelio:
I'm still confused by the O3_CYCLE_TIME = 500 bit.  Can't set just set that to 1, and then let the user invoke ./o3pipeview.py -c 1?  Is there any problem with that?

ccelio:
The default CYCLE_TIME of o3pipeview.py is 1000, so I feel like we should either go with 1 or 1000, if there's no reason either of those doesn't work.

PR #33, opened 2017-10-03
Fix broken links

muojp:
Some links on riscv.org were broken due to website changes.


ccelio:
Thanks for catching and fixing this!

PR #34, opened 2017-10-17
Fixes to get BOOM working in project-template and Firesim

zhemao:
Add xcpt_ae_if to the MicroOp bundle to fix compilation error.

Disable pipeline hang assertion by default.


ccelio:
Thanks for working on this and finding this issue. I'm going to close this as the method I used to fix this is to let the `csr_stall` signal reset the watchdog timer so WFI stalling does not increment the time out. 

Going forward, it'd be nice to have the watchdog timer set by an MMIO chicken bit.

PR #53, opened 2017-12-19
Eliminates a bit of code duplication with Rocket

pentin-as:
I do not like that when in the current versions of rocket-chip RocketCoreplex is changed Boom stops compiling until BoomCoreplex is changed (for example, https://github.com/ucb-bar/riscv-boom/pull/50/commits/08edf16870d7e806f74f4d9e6b513f6dbef7d8cc).

I want, if possible, to inherit BoomCoreplex and BoomTile from Rocket. So far I've done some simpler things and wanted to ask if it makes sense to move in this direction?

I also wanted to let you combine the repositories boom-template and riscv-boom. Now there is riscv-boom, which without rocket-chip does not compile and boom-template, which contains only a set of scripts. This separation is beneficial if you want to build a boom with some other processor other than rocket-chip. Or if instead of Rocket you want to use stubs for tests. Otherwise it seems to me superfluous. Maybe it's better to move the scripts, vsim and the rocket-chip subrepository from boom-template to riscv-boom, and remove boom-template?

In this merge request, Frontend and StratchpadSlavePort from Rocket are inherited.


ccelio:
Thanks for the request. I'll merge it in once I get to the previous pull request, which may be a day or three.

Going forward, let me give you some of the motivation so you understand what I'm thinking.

Of course, reducing code duplication is nice for many reasons, including easier updates. 

A future goal however is to implement a BOOM-specific icache/frontend, which means there's a limit to how much code I can inherit from rocket and that I will instead have to duplicate/rewrite that code (e.g., expanding the interface into Frontend to add additional signals).

> I also wanted to let you combine the repositories boom-template and riscv-boom.

We have been trying out different organizations, and the "project-template" setup is what we have found brings us the most success at Berkeley. The idea is that new projects, tape-outs, and FPGA setups should take boom-template and fork it (add their own scripts, their own bootroms, add in their own git submodules). 

For example, a tape-out of BOOM involves taking the "ExampleBoomSystem" and (for example) wrapping it in a serializing interface, a pad-frame, and other project-specific IP blocks. So while boom-template provides a stand-alone "simulated boom" (which magically talks to the AXI and debug ports) it doesn't provide enough IP for a tape-out, or an FPGA system, which is where others come in and add their IP to their "thin" fork of boom-template. The SiFive Freedom FPGA repo is an example of this setup (https://github.com/sifive/freedom). 

In the past we've tried to use fewer repos, but we found that it made it far more difficult once you had to do something slightly different for a tapeout/fpga/experiment, you then had to fork everything and it was even harder to match upstream. There are quite a few other people right now that use riscv-boom, but NOT boom-template, to build their own boom-based systems.

PR #59, opened 2018-02-07
It was fixed that if the reset of 1 or 2 measures could cause errors

pentin-as:
It was fixed that if the reset of 1 or 2 measures could cause errors: "lsu.scala: 498 assert" or "rob.scala: 579 assert".
+ fixed a bug in unused functionality in Branch Predictor.


ccelio:
Thank you for these fixes.

PR #72, opened 2018-04-20
icache: vaddr to index the cache for lookups and refills

ss2783:
Currently, the icache uses the physical address to compute the tag
index for tag checks and the refills which restricted the ways to
be upto 4KB. This commit allows for ways to be larger than 4KB.
NOTE: The scratchpad behavior should be mostly not affected.

### DIFF
--- a/src/main/scala/ifu/icache.scala
+++ b/src/main/scala/ifu/icache.scala
@@ -75,6 +75,7 @@ class ICachePerfEvents extends Bundle {
 class ICacheBundle(val outer: ICache) extends CoreBundle()(outer.p) {
   val hartid = UInt(INPUT, hartIdLen)
   val req = Decoupled(new ICacheReq).flip
+  val s1_vaddr = UInt(INPUT, vaddrBits) // vaddr delayed one cycle w.r.t. req
   val s1_paddr = UInt(INPUT, paddrBits) // delayed one cycle w.r.t. req
   val s2_vaddr = UInt(INPUT, vaddrBits) // delayed two cycles w.r.t. req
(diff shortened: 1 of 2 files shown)


src/main/scala/ifu/icache.scala
```diff
@@ -110,7 +111,7 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
   val dECC = cacheParams.dataCode
 
   require(isPow2(nSets) && isPow2(nWays))
-  require(!usingVM || pgIdxBits >= untagBits)
+  //require(!usingVM || pgIdxBits >= untagBits)
```
ccelio: delete.

src/main/scala/ifu/icache.scala
```diff
@@ -154,8 +155,9 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
   val hint_outstanding = RegInit(false.B)
   val s2_miss = s2_valid && !s2_hit && !io.s2_kill && !RegNext(refill_valid)
   val refill_addr = RegEnable(io.s1_paddr, s1_valid && !(refill_valid || s2_miss))
```
ccelio: rename to `refill_paddr`.

ccelio (changes requested):
Awesome!

aswaterman:
I suspect this is incorrect for the scratchpad functionality.

ccelio:
Thanks for the review and heads up @aswaterman. The scratchpad is vestigial for BOOM, so hopefully this shouldn't be an issue.

ccelio (approved):

PR #73, opened 2018-04-21
generate a file for the elaborated boom config

ss2783:
This commit adds a file with the .core.config extension which saves
the same information as what was being printed out to STDOUT.

### DIFF
--- a/src/main/scala/bpu/tage.scala
+++ b/src/main/scala/bpu/tage.scala
@@ -125,9 +125,6 @@ class TageBrPredictor(
       table_sizes(i) * tag_sizes(i) + ubit_sz + cntr_sz
    }).reduce(_+_)
 
-   println ("\tBuilding " + (size_in_bits/8/1024.0) + " kB TAGE Predictor ("
-      + (size_in_bits/1024) + " Kbits) (max history length: " + history_lengths.max + " bits)")
-
    require (num_tables == table_sizes.size)
    require (num_tables == history_lengths.size)
    require (num_tables == tag_sizes.size)
@@ -206,6 +203,7 @@ class TageBrPredictor(
    val bp1_idxs = Wire(Vec(num_tables, UInt()))
    val bp1_tags = Wire(Vec(num_tables, UInt()))
 
+   val tables_str = new StringBuilder
    val tables = for (i <- 0 until num_tables) yield
    {
       val table = Module(new TageTable(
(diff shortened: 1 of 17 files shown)


ss2783:
There might be many things that are not scala idiomatic

Mostly works but I can't get rid of some extra empty braces in the execution unit prints. Here is an example:

<snip>
    -== Quad Issue ==-

     ExeUnit--
       - Mem
     ExeUnit--
       - ALU
       - Mul (pipelined)
       - Div
     ExeUnit--
       - ALU
       - IFPU (for read port access)
()         <------ can't get rid of this
     ExeUnit--
       - FPU (Latency: 4)
       - FDiv/FSqrt
       - FPIU (writes to Integer RF)
     ExeUnit--
       - IntToFP
<snio>

ccelio:
Aren't you suppose to be working on your thesis??

aswaterman:
Or busy leapfrogging from company to company

ss2783:
🙌Thanks to RISC-V generated opportunities...juggling both 😀

src/main/scala/exu/core.scala
```diff
@@ -222,44 +226,42 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
-   println("   Load/Store Unit Size  : " + NUM_LSU_ENTRIES + "/" + NUM_LSU_ENTRIES)
-   println("   Num Int Phys Registers: " + numIntPhysRegs)
-   println("   Num FP  Phys Registers: " + numFpPhysRegs)
-   println("   Max Branch Count      : " + MAX_BR_COUNT)
-   println("   BTB Size              : " +
+
+   override def toString: String =
+   ( exe_units_str + "\n"
+   + fp_pipeline_str + "\n"
+   + rob_str + "\n"
```
ccelio, line 235: This is infuriating that rob.toString (and other similar modules) throws a `java.util.NoSuchElementException: head of empty list`. :(

I would like to think there is a way to do this without polluting the code with a bunch of forward declares.

src/main/scala/bpu/tage.scala
```diff
@@ -222,9 +220,15 @@ class TageBrPredictor(
...
       if (i > 0) require(history_lengths(i) > history_lengths(i-1))
 
+      tables_str.append(table.toString + "\n")
       table
    }
 
+   override def toString: String =
+      ("  Building " + (size_in_bits/8/1024.0) + " kB TAGE Predictor (" +
+      + (size_in_bits/1024) + " Kbits) (max history length: " + history_lengths.max + " bits)\n"
+      + tables_str.toString)
```
ccelio, line 230: You can instead use `tables.mkString("\n")` to do this instead. https://www.artima.com/pins1ed/next-steps-in-scala.html

ccelio (approved):
Thanks for doing this work. I'm going to clean it up a bit and merge it in now.

ccelio:
Huh, it doesn't recognize I rebased this. Anyways, PR merged!

src/main/scala/exu/core.scala
ccelio: This looks like a Chisel bug. If I print out the core.toString everything works, but the ElaborationArtefacts() pukes on it, and it's on a chisel3.internal.naming.NamingStack.pop_return_context call. I'll escalate this elsewhere.

ccelio: Oh, and I think this may be a `chiselName` macro error. Fun times.

PR #86, opened 2018-05-30
Valid predictors

pentin-as:
1. Fixed that MegaBoomConfig was not compiled
The requirement was not fulfilled:
https://github.com/ucb-bar/riscv-boom/blob/aaf42c950cd1620d7b6c4e6cc22b43aa306f3739/src/main/scala/bpu/dense-btb.scala#L100

2. Fixed that bim, dense-btb, tage could make invalid predictions
See issue:
https://github.com/ucb-bar/riscv-boom/issues/85


ccelio:
Very sorry I missed this. @jerryz123 this is good to merge.

Issue #42, opened 2017-11-09
Load-reserve takes erroneous page fault on TLB xcpt.store signal.


ccelio:
Load-reserve thinks it's a store. Bad load-reserve.

PR #95, opened 2018-07-25
[lsu] Fixed load-reserved exceptions

jerryz123:
* Enables Linux boot

### DIFF
--- a/src/main/scala/lsu/lsu.scala
+++ b/src/main/scala/lsu/lsu.scala
@@ -398,16 +398,28 @@ class LoadStoreUnit(pl_width: Int)(implicit p: Parameters, edge: freechips.rocke
 
    // exceptions
    val ma_ld = io.exe_resp.valid && io.exe_resp.bits.mxcpt.valid && exe_tlb_uop.is_load
-   val pf_ld = dtlb.io.req.valid && dtlb.io.resp.pf.ld && exe_tlb_uop.is_load
+   val pf_ld = dtlb.io.req.valid && dtlb.io.resp.pf.ld && (exe_tlb_uop.is_load || exe_tlb_uop.is_amo)
    val pf_st = dtlb.io.req.valid && dtlb.io.resp.pf.st && exe_tlb_uop.is_store
-   val ae_ld = dtlb.io.req.valid && dtlb.io.resp.ae.ld && exe_tlb_uop.is_load
+   val ae_ld = dtlb.io.req.valid && dtlb.io.resp.ae.ld && (exe_tlb_uop.is_load || exe_tlb_uop.is_amo)
    val ae_st = dtlb.io.req.valid && dtlb.io.resp.ae.st && exe_tlb_uop.is_store
    // TODO check for xcpt_if and verify that never happens on non-speculative instructions.
-   val mem_xcpt_valid = Reg(next=((dtlb.io.req.valid && (pf_ld || pf_st || ae_ld || ae_st)) ||
+   val mem_xcpt_valid = Reg(next=((pf_ld || pf_st || ae_ld || ae_st) ||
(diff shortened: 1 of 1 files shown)


src/main/scala/lsu/lsu.scala
```diff
@@ -1260,6 +1260,10 @@ class LoadStoreUnit(pl_width: Int)(implicit p: Parameters, edge: freechips.rocke
       {
          stq_tail := stq_commit_head
 
+         when (!stq_entry_val(stq_execute_head)) {
```
ccelio: Interesting catch! What is the scenario where the `execute_head` is running ahead of the tail? This looks like a bug with the `execute_head` advance logic, and this code should be an assert to catch it, not a hack to paper over it.

jerryz123: image

Here entry 7 contains a faulting LR instruction. You can see that the execute head runs past the tail when the tail is set to the commit_head. 

One solution I see is to increment the head past a faulting entry. That would fix this test case, but it might have other side effects

ccelio: I suspect the correct solution is to nack a faulting AMO to prevent the stq_execute_head from advancing (or prevent it from executing all together). The `stq_entry_val` check leaves me incredibly nervous.

jerryz123: I'll try that. 

What's strange to me is that this bug shouldn't just be limited to LR instructions. Shouldn't any faulting store trigger the same failure?

ccelio: I think it's a larger bug reported in (https://github.com/ucb-bar/riscv-boom/issues/42); should the LR ever be placed into the STQ? If it is in the STQ then I assume there's a control signal somewhere that thinks it is a load and thus doesn't push the store nack further up the chain.

davidbiancolin: +1 On adding an assert regardless of the fix. How long after that ^ behavior is an assertion throw and what kind of assertion is it?

jerryz123: There already is an assertion to catch the head going out of sync, and it triggers immediately at the end of the ^ behavior.

davidbiancolin: Cool

jerryz123:
Here I do not send a store to dcache if the store generated an exception. This requires using 1 extra bit of state per store entry to track if that entry generated an exception.

@ccelio I tried implementing what you had in #42, but that fix seems to require more changes throughout the pipeline.

ccelio:
Thanks @jerryz123, I think this is fine. It explicitly notes that the AMO has suffered an exception and sending it to the D$ is blocked off.

Just for the record, here are my thoughts on this problem:

First, `is_amo` is being misused. In decode.scala, it is incorrectly being used to say "is this instruction in the A extension?" This causes LR to be marked as an "amo" when it isn't. I think LR should actually be a new micro-op: it unfortunately can't be a load uop because the address mode is different, but I don't like the confusion of the LR using the existing AMO-AGEN micro-op. Then, `is_amo` can be used exclusively to describe uops that atomically read and write memory.

I don't think you need to add an extra "exception" bit to the LSU queues. Instead, you only need to stop the store/AMO from executing, so on detection of an exception, I would set the `stq_executed` bit to denote that the store/amo does not need to be executed. There's a one cycle shadow between detecting the exception and setting the `stq_executed` bit, so the `can_store_fire` will need to be suppressed when the `mem_xcpt` is detected for an AMO that may try to proceed.

But if your thing boots linux, then by all means merge it in. I don't see any holes/further issues with it.

jerryz123:
Ok I'll merge this in now, but I'll try your suggestion when I get around to working on the general LR issue.

PR #97, opened 2018-07-29
chisel3 syntax and support for riscv compressed instructions

codelec:
-  edits made throughout the project to no longer use compatibility mode offered by chisel3
-  RVC Support 
rvc-flow


jerryz123:
I'm not sure how easy this would be to implement, but it would be very convenient if RVC support was toggleable as a parameter, instead of enabled by default.

pranith:
It would be great if you could submit the RVC support in a separate pull request and the chisel3 changes in a separate pull.

Issue #101, opened 2018-08-07
False page-fault and access exceptions.


jerryz123:
We are seeing a bug with BOOM running user programs in Linux in our Firesim testing. 

```
[    0.996000] xargs[57]: unhandled signal 11 code 0x1 at 0x0000000000000038 in libc-2.26.so[2000024000+13e000]
[    0.996000] CPU: 0 PID: 57 Comm: xargs Not tainted 4.15.0-rc6-31580-g9c3074b5c2cd-dirty #21
[    0.996000] sepc: 0000002000093804 ra : 00000020000937d0 sp : 0000003fffc1da80
[    0.996000]  gp : 00000000000d54c8 tp : 00000020000236d0 t0 : 0000000000000659
[    0.996000]  t1 : 0000002000094958 t2 : ffffffffffffffff s0 : 0000000000000000
[    1.000000]  s1 : 0000000000000888 a0 : 0000000000000000 a1 : 00000000000de270
[    1.000000]  a2 : 00000000000df270 a3 : fffffffffffff7ff a4 : 0000002000092a6c
[    1.000000]  a5 : 00000000000de270 a6 : 00000000000000ff a7 : 0000000000000104
[    1.000000]  s2 : 00000020001648f0 s3 : 0000000000000bd0 s4 : 0000002000164c38
[    1.000000]  s5 : 0000000000000000 s6 : 0000000000000000 s7 : 00000000000b75d8
[    1.000000]  s8 : 00000000000d51b8 s9 : 00000000000077f5 s10: 0000000000000000
[    1.000000]  s11: 0000000000000040 t3 : 000000200009106c t4 : 0000000000000002
[    1.000000]  t5 : 0000002000027368 t6 : 0000000000000000
[    1.000000] sstatus: 8000000200006020 sbadaddr: 0000000000000038 scause: 000000000000000d
```

- The unhandled signals are always 7 (SIGBUS) or 11 (SIGSEGV). 
- The error is extremely sensitive to the state of the machine
- sepc does not always point to a memory instruction, (although I'm not 100% certain I'm reading the right instruction from the sepc)
- sbadaddr is an unreasonable address

Basically, BOOM is generating false access and pagefault exceptions for user code. I haven't been able to reproduce this issue in a smaller test case. Actually, I'm not always able to reproduce this issue at all. 

What's strange to me is what could be setting sbadaddr to such strange values. It seems like an invalid request is being sent through the LSU somehow, but I don't see where this could be possible.

[...]

PR #102, opened 2018-08-08
[ifu] Disallow aliasing in Icache

jerryz123:
This is a workaround for #101.


aswaterman:
When you figure out the actual bug, can you write an explanation on this thread and @-mention me?

Although I'm not currently building any I$ that alias, I was pretty sure the Rocket design is correct in that respect.

jerryz123:
@aswaterman 

This is probably a specific bug with BOOM's banked I$, rather than the general Rocket design.

ccelio:
@aswaterman I believe the bug is related to computing the virtual and physical tag and idx bits. Those are hardcoded in rocket under the VIPT assumption, iirc, and need to be overridden by the consumer.

PR #112, opened 2018-09-07
[core] Added optional tracing for retired instructions

jerryz123:
The retired instruction trace is generated by storing fetched PCs and insns in the ROB. This is not ideal for  actual hardware, but is acceptable for debugging simulations on FPGA.


ccelio:
Somewhat off-topic, but I think Chisel could really use an annotation to mark signals that should be marked as "unsynthesized" in the Verilog. And if the proper flags during synthesis aren't set, an error is thrown.

colinschmidt:
Sounds like an interesting chisel/firrlt feature request...

PR #124, opened 2018-11-03
PNR + Secure MSHR WIP

bkorpan:
Added point-of-no-return ROB head and provisions to kill inflight speculative cachefills. Currently broken; work in progress.

### DIFF
--- a/src/main/scala/common/parameters.scala
+++ b/src/main/scala/common/parameters.scala
@@ -230,8 +230,8 @@ trait HasBoomCoreParameters extends freechips.rocketchip.tile.HasCoreParameters
 
    //************************************
    // Implicitly calculated constants
-   val NUM_ROB_ROWS      = NUM_ROB_ENTRIES/decodeWidth
-   val ROB_ADDR_SZ       = log2Up(NUM_ROB_ENTRIES)
+   val NUM_ROB_ROWS      = NUM_ROB_ENTRIES/fetchWidth
+   val ROB_ADDR_SZ       = log2Up(NUM_ROB_ENTRIES) + 1
    // the f-registers are mapped into the space above the x-registers
    val LOGICAL_REG_COUNT = if (usingFPU) 64 else 32
    val LREG_SZ           = log2Up(LOGICAL_REG_COUNT)
--- a/src/main/scala/exu/rob.scala
+++ b/src/main/scala/exu/rob.scala
@@ -29,11 +29,11 @@ package boom.exu
 
 import Chisel._
 import chisel3.experimental.chiselName
-import scala.math.ceil
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.util.Str
 import boom.common._
 import boom.util._
+import chisel3.experimental.dontTouch
 
 class RobIo(
    val machine_width: Int,
@@ -76,6 +76,9 @@ class RobIo(
    // tell the LSU that the head of the ROB is a load
    // (some loads can only execute once they are at the head of the ROB).
    val com_load_is_at_rob_head = Bool(OUTPUT)
+
+   // The LSU needs the PNR head to wake up spec buffer entries.
+   val rob_pnr_head = UInt(OUTPUT, ROB_ADDR_SZ)
 
    // Communicate exceptions to the CSRFile
    val com_xcpt = Valid(new CommitExceptionSignals())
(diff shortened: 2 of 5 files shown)


src/main/scala/common/parameters.scala
```diff
@@ -230,8 +230,8 @@ trait HasBoomCoreParameters extends freechips.rocketchip.tile.HasCoreParameters
 
    //************************************
    // Implicitly calculated constants
-   val NUM_ROB_ROWS      = NUM_ROB_ENTRIES/decodeWidth
-   val ROB_ADDR_SZ       = log2Up(NUM_ROB_ENTRIES)
+   val NUM_ROB_ROWS      = NUM_ROB_ENTRIES/fetchWidth
```
ccelio, line 233: Why the change to fetchWidth? You want the ROB structure completely decoupled from the front-end/fetchWidth.

src/main/scala/exu/rob.scala
```diff
@@ -204,8 +208,10 @@ class Rob(
 
 
    //commit entries at the head, and unwind exceptions from the tail
-   val rob_head = Reg(init = UInt(0, log2Up(num_rob_rows)))
-   val rob_tail = Reg(init = UInt(0, log2Up(num_rob_rows)))
+   val rob_head = Reg(init = UInt(0, rob_row_addr_sz))
+   val rob_tail = Reg(init = UInt(0, rob_row_addr_sz))
+   val rob_pnr_head = Reg(init = UInt(0, rob_addr_sz))
```
ccelio, line 213: For brevity, I recommend calling this "rob_pnr". So there's "*head, *tail, and *pnr".

src/main/scala/exu/rob.scala
```diff
@@ -677,23 +689,36 @@ class Rob(
...
+   // ROB Point-of-No-Return (PNR) Head Logic
+
+   // The PNR head operates on the granularity of individual ROB entries, rather than ROB rows.
+   // Additionally, it is able to jump ahead by an arbitrary number of entries each cycle.
+   // This allows waiting speculative stores in the SAQ or speculative caches fills in the MSHR to commit more quickly.
+   // It points to the oldest "unsafe" instruction - if no instructions are unsafe, it is set to the tail.
+
+   rob_pnr_head := Mux(rob_unsafe.asUInt.orR,
+                       PriorityEncoder((rob_unsafe.asUInt << (UInt(num_rob_entries) - rob_head(rob_row_addr_sz-2,0) * UInt(width)) | rob_unsafe.asUInt >> (rob_head(rob_row_addr_sz-2,0) * UInt(width)))(num_rob_entries-1,0)) + rob_head * UInt(width),
+                       rob_tail * UInt(width))
```
ccelio, line 705: Use the new "width.U" syntax.

src/main/scala/common/parameters.scala
```diff
@@ -230,8 +230,8 @@ trait HasBoomCoreParameters extends freechips.rocketchip.tile.HasCoreParameters
...
+   val ROB_ADDR_SZ       = log2Up(NUM_ROB_ENTRIES) + 1
```
ccelio, line 234: Why the +1? Are you embedding extra information in the pointer? Consider coding style/naming convention to make this easier for others to intuit your reasoning. For example, I'm a fan of rocket-chip's "VaddrExtended". Instead of doing "vaddr+1" everywhere, it becomes more clear that something "magic" is happening with the vaddr (it's compressing XLEN-vaddr bits down to 1 bit).

src/main/scala/exu/rob.scala
```diff
@@ -440,6 +450,12 @@ class Rob(
          rob_val(rob_head) := false.B
       }
 
+      // -----------------------------------------------
+      // Is it safe to move the PNR head past this ROB entry?
+      for (i <- 0 until num_rob_rows) {
+         rob_unsafe(i*width + w) := rob_val(i) && rob_bsy(i) && (rob_uop(i).is_br_or_jmp || rob_uop(i).is_load || rob_uop(i).is_store) || rob_exception(i)
```
ccelio, line 456: I would recommend against allowing the rob_pnr to jump arbitrary lengths. First, matching the ROB enqueue bandwidth of one row/cycle is likely to be perfectly acceptable. Also consider what hardware you are describing here.

For my proposal, the unsafe bits aren't needed. Just have the rob_pnr scan one row at a time just like the rob_head looking at rob_val, rob_bsy, and uopc (although reading uopc is also probably very expensive, so maybe add an unsafe_type bit and set it in enqueue). This is cheap and a fairly quick scan.

In this code you are adding combinational logic for every ROB entry to be updated every cycle. But the more problematic section of RTL is the PriorityEncode which scans across all nRobEntries bits -- this may become a critical path. 
 
Comparing the two approaches would be a fun perf. study to perform, just to see if the PNR bandwidth is fine if it only increments one (or two) row(s) at a time.

ccelio (commented):
One final comment; I recommend breaking apart different "updates" into multiple PRs. In this case, separate PNR and the SecureMSHR. Makes it easier to review and if one part is a WIP or broken you can easily undo/separate it from the other pieces.

PR #126, opened 2018-11-09
[util] Generalize killable uop queues

jerryz123:
Create a general module for queues containing BOOM MicroOps which may be killed by branch resolution.


ccelio (approved):
lgtm.

PR #131, opened 2018-11-29
[dec] Fix CSR decode

jerryz123:
This matches rocket-chip. Resolves broken `rv64mi-p-illegal` test.

https://github.com/freechipsproject/rocket-chip/blob/e03605a934ee1b276ef9b743fa3f5d1c009788ed/src/main/scala/rocket/RocketCore.scala#L251

@ccelio I'm curious why was this ever set to `>=`?


ccelio:
Because that is what was there before: https://github.com/freechipsproject/rocket-chip/blame/19b569e69a6c9f2b3796ad70892821718055f9c8/src/main/scala/rocket/RocketCore.scala#L202

And changed for "QoR reasons":
https://github.com/freechipsproject/rocket-chip/blame/e03605a934ee1b276ef9b743fa3f5d1c009788ed/src/main/scala/rocket/RocketCore.scala#L251

Issue #120, opened 2018-10-26
Chisel3 support


jerryz123:
We should move BOOM to Chisel3. #97 addresses some of this, but the Chisel3 changes need to be separate from RVC changes.

PR #132, opened 2018-12-06
BOOM Chisel3 update

abejgonzalez:
This is porting the BOOM codebase to use proper Chisel3 syntax.

Tested with run-asm-tests, run-bmark-tests on SmallBoomConfig and BoomConfig.

Note: RocketChip although using the "import Chisel._" Chisel2 import, is actually using Chisel3 compatibility mode.

### DIFF
--- a/src/main/scala/bpu/2bc-table.scala
+++ b/src/main/scala/bpu/2bc-table.scala
@@ -41,13 +41,14 @@
 
 package boom.bpu
 
-import Chisel._
+import chisel3._
+import chisel3.util._
 import boom.util.SeqMem1rwTransformable
 
 
 class UpdateEntry(fetch_width: Int, index_sz: Int) extends Bundle
 {
-   val index            = UInt(width = index_sz)
+   val index            = UInt(index_sz.W)
    val executed         = Vec(fetch_width, Bool())
    val takens           = Vec(fetch_width, Bool())
    // Was there a misprediction? If yes, we need to read the h-tables.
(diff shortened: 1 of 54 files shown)


abejgonzalez:
Addresses #120

jerryz123:
LGTM, I think we should just verify Linux on FireSim, and then we can merge

sagark:
If you guys are doing that, can you also PR the bumped BOOM to FireSim dev branch and share the updated fpga images in FireSim?

jerryz123:
Let's merge this, merge the PMP stuff, then bump

jackkoenig:
I would suggest generating a few BOOM configs and then using LEC to prove that this doesn't change the functionality, just to check for unexpected changes in semantics. There aren't supposed to be many but there are a few (in particular, in compatibility mode `:=` is bidirectional, it is monodirectioned in `import chisel3._`)

abejgonzalez:
Chisel3 port allows Linux to boot in FireSim (fireboom-singlecore-nic-ddr3-llc4mb).

Additionally, there are changes in the commits to address := by switching it to a bulk connection <>. Or connecting individual components together.
> (in particular, in compatibility mode `:=` is bidirectional, it is monodirectioned in `import chisel3._`)

In the processes of running LEC checks.

abejgonzalez:
LEC using Formality passes on module BoomCore for both the SmallBoomConfig and default BoomConfig. Any final concerns or comments? Or I will go ahead and merge!

jerryz123:
Seems good to me, what was the issue with only 49% equivalence you were seeing earlier?

abejgonzalez:
One thing, was that I was getting a SIGFPE error when running on BoomTile so I wasn't able to get matching results when comparing that to each other. For BoomCore, after running the tool to ignore "signed to unsigned assignment", I also didn't see the 49% anymore. I dont remember exactly the steps that were taken to resolve the issue since I was trying multiple different things but it looks like things are equivalent with mainly unread points (signals that dont drive anything) not being compared. Let me see if I can recreate it.

abejgonzalez:
So it looks like the 49% was gotten from the matching done on BoomTile for BoomConfig. However, I am not able to diagnose the issue due to the crashing after this point. Looking at the BoomTile code changes the only main difference is the additions of the DontCares (and these signals are not being assigned in the rest of the tile). So I think it is safe to assume it is fine.

jerryz123:
Yeah it's fine, I was just curious what might be causing the divergence

abejgonzalez:
Its seems to be from the BoomTile side but with Formality crashing I can't get more info out. :(

PR #136, opened 2018-12-12
Adjusted syntax according to scalastyle-config

abejgonzalez:
Updated most of the syntax to match the scalastyle config that is currently present. This is part of the work to get Travis CI up and running with no errors (this will be one of the things that is run during Travis).


abejgonzalez:
A quick thought: Do we want to have a different scalastyle config (shorter line width, etc)? Maybe that can be a question for later? This is just to get us started.

abejgonzalez:
@ccelio Im assuming we can change the headers to say 2018 instead of 2015. Is that fine?

ccelio:
120 line width comes from Java. I think it's a fine size for two-wide on modern monitors.

I can't speak to what to do with copyright headers; I guess update them?

I wouldn't blindly sed update 2015 to 2018 though. Looks like you hit some other uses.

abejgonzalez:
What do you mean by "looks like you hit some other uses". As for the headers, I thought it would be fine to update whenever the file has changed (so for the Chisel3 update most files were changed hence the 2015->2018 change).

sagark:
AFAIK these headers should be [first written year - now], you’re not
supposed to just change the year to the current year.

src/main/scala/bpu/brpredictor.scala
```diff
@@ -1,14 +1,15 @@
...
 // All Rights Reserved. See LICENSE for license details.
 //------------------------------------------------------------------------------
+
 //------------------------------------------------------------------------------
 // RISCV Branch Predictor (abstract class)
 //------------------------------------------------------------------------------
 //------------------------------------------------------------------------------
 //
 // Christopher Celio
-// 2015 Oct 12
```
ccelio, line 11: Sed script collateral damage.

src/main/scala/bpu/gshare.scala
```diff
@@ -1,14 +1,15 @@
...
 // All Rights Reserved. See LICENSE for license details.
 //------------------------------------------------------------------------------
+
 //------------------------------------------------------------------------------
 // RISCV GShare Branch Predictor
 //------------------------------------------------------------------------------
 //------------------------------------------------------------------------------
 //
 // Christopher Celio
-// 2015 Apr 28
```
ccelio, line 11: Another victim. F.

abejgonzalez:
Ahh, ok I will change the code back to have a range and mention authors

abejgonzalez:
Ok, I moved the authors to the top headers and fixed the dates so that they are consistent from before (if the file did not have a date I just specified it from the last known commit (the earliest commit to that file)

abejgonzalez:
Any other thoughts or concerns @ccelio or is this good to merge?

ccelio:
Looks good to me.

PR #137, opened 2018-12-19
Fix MegaBoomConfig by increasing icache associativity

pranith:
There is a requirement that we can only support 64 sets in the icache due to
aliasing bugs. Increase associativity and reduce the sets for icache to fix the
default MegaBoomConfig. This icache config is similar to DefaultBoomConfig.


abejgonzalez:
Looks good to me

pranith:
Fixed the conflict.

Issue #92, opened 2018-06-21
Travis CI?


jerryz123:
Has there been any work done towards enabling Travis for BOOM, similar to what is done for RocketChip?

ccelio:
Nope. Feel free to take a stab at it though.

There are a couple of challenges here. 

First, ideally the Travis would gate commits to riscv-boom, but riscv-boom is not a standalone repository. So any changes that require a synchronized update across dependent repositories (boom-template, rocket-chip, riscv-tools) would not work for this type of CI flow.

There is also is the run/build time. Travis often timed out on rocket, and rocket is relatively simply and runs much faster than boom. There's also the hassle of trying to get it to cache the riscv-tools to improve run time, instead of rebuilding it every time.

Watching from afar, it seemed like a headache for freechips/sifive to deal with, so I never gave it a go.

aswaterman:
Travis nodes are also limited to 8 GiB of memory.

ccelio:
Ouch, thanks for the additional info @aswaterman .

jerryz123:
Thanks for the info, I'm trying to figure out if this is feasible before I try.

> First, ideally the Travis would gate commits to riscv-boom, but riscv-boom is not a standalone repository.

Couldn't we make the Travis script fetch some stable release of boom-template and its submodules, and the current commit of BOOM?

> There is also is the run/build time. Travis often timed out on rocket, 

It seems like Travis has no timeout limit, assuming every job outputs something every 10 minutes. 

> Travis nodes are also limited to 8 GiB of memory.

Is 8 GB of memory not enough for the build? I'm a little more concerned about the disk space. which is capped at 22 GB for the Ubuntu 12.04 image. My boom-template folder right now is 20 GB.

ccelio:
The problem is when you need to make a change across rocket-chip+riscv-boom+boom-template, there won't yet exist the proper boom-template to test against, since the leaf repos have to be updated first. The only thing you can do is pull out your sudo permissions and override the Travis failure gating off commits.

I thought I saw that boom takes ~13 GB of VM during Chisel compile, but maybe I'm wrong about that. Your folder should only be ~1 GB, unless you are storing *.vpd files or you keep around the riscv-tools build.

jackkoenig:
@ccelio, is the ~13 GB figure Chisel proper or kind of Chisel and FIRRTL? If it's the former, what do you see for FIRRTL?

ccelio:
[...]

PR #138, opened 2018-12-21
Travis Integration

abejgonzalez:
This is the first step to getting Travis setup with Boom. This makes an initial build of the default BoomConfig and tests it against Scalastyle checks, regression tests, and benchmark tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,97 @@
+git:
+    submodules: false # dont clone submodules
+
+# default language
+language: scala
+
+# need sudo to build tools/verilator
+sudo: required
+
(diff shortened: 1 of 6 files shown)


abejgonzalez:
Note: This topic was discussed in #92. This uses the master branch of riscv-boom/boom-template to build the project. Additionally, run-asm-tests was unable to run since Travis jobs timeout at 2hrs. Maybe when (https://github.com/freechipsproject/rocket-chip/pull/1665) then asm tests can be faster.

jerryz123:
Can you add the Travis sticker to the Readme?

PR #141, opened 2018-12-29
RVC Support

jerryz123:
In time for 2019, here is work towards RVC support. 

Procedure:
1. FetchController gets a fetchWidth*16b packet from ICache
2. FetchController generates a fetchWidth*32b FetchBundle from the incoming packet
    - 16b trailing bits are kept behind for misaligned 32b instructions
    - RVC/non-RVC instructions are marked here
3. FetchBuffer compacts away unused instructions when en-queuing the FetchBundle
4. Decode uses Rocket's RVCExpander

TODO:
- [x] Verify exception handling
- [x] Parameterize this (allow both RVC and non-RVC configs?)
- [x] Test on benchmarks built with RVC
- [x] Update boom-template README, tools, boom
- [x] Verify branch predictors/RAS functionality
- [x] Verify running RVC pk and bbl
- [x] Verify RVC Linux boot
- [x] Verify on other configurations, with various `fetchWidths`
- [x] Add documentation/comments
- [x] Update README
- [x] Disable printf


src/main/scala/ifu/fetchbuffer.scala
```diff
@@ -117,7 +133,7 @@ class FetchBuffer(num_entries: Int)(implicit p: Parameters) extends BoomModule()
 
    // all enqueuing uops have been compacted.
    // How many incoming uops are there?
-   val popc_enqmask = PopCount(io.enq.bits.mask)
+   val popc_enqmask = PopCount(in_uops.map(_.valid))
```
jerryz123, line 137: @ccelio This change actually breaks simulations where memory is initialized to X instead of RAND (`+define+RANDOMIZE_MEM_INIT`). `in_uops(i).valid` holds `x`s when the frontend fetches from uninitialized memory, and this screws up the `count` register in the fetchbuffer. 

Are there any suggestions for getting around this? Is this worth trying to resolve? Deterministically initialized memory makes comparisons between configurations a lot easier.

jerryz123:
> I find fetchWidth very confusing -- my recommendation is to add/use fetchBytes (which I think is in baseTile) and then compute fetchWidth off of that.

@ccelio I would prefer to actually keep configuring fetchWidth, since rocketchip's CoreParams calculates fetchBytes from fetchWidth. 

https://github.com/freechipsproject/rocket-chip/blob/e3a9ee30f278c8a5cf4160cc72051dd19c30f819/src/main/scala/tile/Core.scala#L44

src/main/scala/common/consts.scala
```diff
@@ -23,7 +23,7 @@ import freechips.rocketchip.util.Str
 trait BOOMDebugConstants
 {
    val DEBUG_PRINTF        = false // use the Chisel printf functionality
-   val COMMIT_LOG_PRINTF   = false // dump commit state, for comparision against ISA sim
+   val COMMIT_LOG_PRINTF   = true // dump commit state, for comparision against ISA sim
```
abejgonzalez: Do we want to have this enabled by default? I personally would prefer this to be off to not pollute the output during sim.

jerryz123: Yep good catch l'll delete it once done debugging

jerryz123:
@abejgonzalez @ccelio This seems to work now. Any last concerns?

Also I'd prefer to just merge this into master now instead of waiting for the CI change to avoid generating a bunch of Merge commits in the history.

README.md
```diff
@@ -90,39 +90,15 @@ will also need to add $RISCV/bin to your $PATH.
...
-   $ cd boom-template/rocket-chip
-   $ git submodule update --init
+   $ git submodule update --init rocket-chip
+   $ cd rocket-chip
+   $ git submodule update --init riscv-tools
    $ cd riscv-tools
    $ git submodule update --init --recursive
-   $ cp build.sh build-rv64g.sh
-   $ vim build-rv64g.sh
+   $ ./build.sh
```
abejgonzalez: I would instead have this section point to the boom-template build-tools script. 

But we have to make sure that boom-template is updated.

docs/sections/InstructionFetch/FetchStage.rst
```diff
@@ -46,8 +46,73 @@ miss has been handled. This is less than ideal for scenarios in which
...
+code size. “RVC" comes with a number of features that are of particular
+interest to micro-architects:
+
+-  32b instructions have no alignment requirement, and may start on a
+   half-word boundary.
+
+-  All 16b instructions map directly into a longer 32b instruction.
+
+BOOM re-uses the front-end design from Rocket, a 5-stage in-order core.
+BOOM then takes instructions returning (the *fetch packet*) from the
```
abejgonzalez, line 64: I think this is slightly weird sentence structure. To me, this sounds better as...

After receiving a fetch packet of instructions from the Rocket front-end, BOOM quickly decodes ...

README.md
jerryz123: I don't see a good reason to keep that script around. The CI would use the original `build-tools.sh` in riscv-tools, and so would the README and docs. We could add to the README information on building rv64g binaries with a rv64gc default toolchain.

I'll add some clarifications here on why we use riscv-tools/build.sh, and what to do if you are building no-rvc BOOM (although I doubt anyone will do this in the future).

abejgonzalez: Since you now build the GC version of riscv-tools, I think the boom-template build script is mainly for convenience (since it just copies those basic commands). I think it would be nice to keep it.

jerryz123: All boom-template/scripts/build-tools.sh does is emit a version of riscv-tools/build.sh that has `--with-arch=rv64imafd` to set the default isa to `rv64g`. I'm down to change it to just call riscv-tools/build.sh

abejgonzalez:
Also need to update the following: 
        riscv-boom/docs/sections/IntroductionOverview/RISCV-ISA.rst
 To mention that RVC is now up and running.

jerryz123:
> Also need to update the following:
> riscv-boom/docs/sections/IntroductionOverview/RISCV-ISA.rst
> To mention that RVC is now up and running.

Thanks, I missed that

jerryz123:
Merging. Note #153 still blocks Fedora boot

PR #142, opened 2019-01-01
fix Generator.scala to emit correct set of tests to be run by BOOM without an FPU

codelec:
disabling fpu should show the correct set of supported extensions in dts

### DIFF
--- a/src/main/scala/common/configs.scala
+++ b/src/main/scala/common/configs.scala
@@ -57,7 +57,7 @@ class DefaultBoomConfig extends Config((site, here, up) => {
 
 class WithoutBoomFPU extends Config((site, here, up) => {
    case BoomTilesKey => up(BoomTilesKey, site) map { r => r.copy(core = r.core.copy(
-      usingFPU = false))
+      fpu = None))
    }
 })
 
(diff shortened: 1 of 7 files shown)


src/main/scala/common/configs.scala
```diff
@@ -57,7 +57,7 @@ class DefaultBoomConfig extends Config((site, here, up) => {
 
 class WithoutBoomFPU extends Config((site, here, up) => {
    case BoomTilesKey => up(BoomTilesKey, site) map { r => r.copy(core = r.core.copy(
-      usingFPU = false))
```
jerryz123, line 60: Can we keep the flag as usingFPU instead of fpu for consistency with rocketchip?

jerryz123 (changes requested):

src/main/scala/common/configs.scala
codelec: setting fpu to None does disable `usingFPU` https://github.com/freechipsproject/rocket-chip/blob/master/src/main/scala/tile/Core.scala#L54

jerryz123 (approved):

jerryz123:
Thank you @codelec

PR #143, opened 2019-01-02
remove cloneType in favour of autoclonetype 

codelec:
* remove cloneType in favour of autoclonetype 
* explicitly disabling strict chisel3 checking seems to be no longer required

### DIFF
--- a/src/main/scala/util/elastic-sram.scala
+++ b/src/main/scala/util/elastic-sram.scala
@@ -10,12 +10,11 @@ package boom.util
 import chisel3._
 import chisel3.util._
 
-class ESramWritePort[T <: Data](idx_sz: Int, gen: T) extends Bundle
+class ESramWritePort[T <: Data](val idx_sz: Int, val gen: T) extends Bundle
 {
    val idx = UInt(idx_sz.W)
    val data = gen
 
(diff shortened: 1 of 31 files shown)


abejgonzalez:
Looks good to me. As for the strict chisel3 checking in compileOptions, it seems like that was mainly for chisel2 compatibility mode. Now that there are DontCares in every spot that is unspecified (so far as the CI checks show) that is not needed. Thanks!

src/main/scala/util/elastic-sram.scala
```diff
@@ -10,12 +10,11 @@ package boom.util
 import chisel3._
 import chisel3.util._
 
-class ESramWritePort[T <: Data](idx_sz: Int, gen: T) extends Bundle
+class ESramWritePort[T <: Data](val idx_sz: Int, val gen: T) extends Bundle
```
jackkoenig, line 13: This one will run afoul of a gotcha. By making these `val`s, they become public fields of the class, which means that `val gen` will be a public field of type `Data`, thus making it part of the Bundle. You should either make `gen` a `private val` or a function `val gen: () => T`

abejgonzalez: Oh. I was unaware of that. Is there a difference in making it a function or private val. Both are functionally equivalent correct?

PR #146, opened 2019-01-06
[ci] Pull prebuilt release of riscv-tools

jerryz123:


jerryz123:
Closed because there are better ways to do this

PR #160, opened 2019-01-22
[lsu] Fix speculative load wakeups

### LINKED ISSUE #153: Bug with 4-cycle loaduse
There's a bug with 4-cycle load-use. Here are two runs in which it appeared. The first is a dual-core config booting buildroot Linux. The second is a single-core config booting Fedora.

```
id: 1190, module: IssueSlot_4, path: FireBoomNoNIC.tile_1.core.issue_units_0.slots_3]
Assertion failed
    at issue_slot.scala:214 assert (!slot_p1_poisoned)
    at cycle: 2142042185
```
```
[id: 601, module: IssueSlot_4, path: FireBoomNoNIC.tile.core.issue_units_0.slots_2]
Assertion failed
    at issue_slot.scala:216 assert (!slot_p1_poisoned)
 at cycle: 195769823414
```

jerryz123:
Loads to zero register should not speculatively wakeup other uops,

Resolves #153

### DIFF
--- a/src/main/scala/exu/core.scala
+++ b/src/main/scala/exu/core.scala
@@ -159,7 +159,10 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
 
    // Load/Store Unit & ExeUnits
    exe_units.memory_unit.io.lsu_io <> lsu.io
+
+   // TODO: Generate this in lsu
    val sxt_ldMiss = Wire(Bool())
+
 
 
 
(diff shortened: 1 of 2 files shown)


abejgonzalez (approved):

ccelio:
I don't know if my code review comment was lost, but you should also add an assertion to check that the issue window never receives a wakeup for x0.

jerryz123:
> I don't know if my code review comment was lost, but you should also add an assertion to check that the issue window never receives a wakeup for x0.

I believe operations on x0 still go through execute and writeback, so in general it should be ok to still receive "wakeups" for x0? For normal wakeups the present bit should be set for x0 always anyways so this doesn't affect the issue slot entry at all.

The issue was from the ldspec wakeups and ldmiss signals messing up the poison bit.

ccelio:
Sorry, I meant to say to check that you don't receive speculative load wakeups for x0. Or said another way, you want to verify you never have a _poisoned_ rs1/rs2 for when they equal x0.

PR #163, opened 2019-01-23
Add asserts to catch specld wakeups to x0

jerryz123:
Addresses remark in #160

### DIFF
--- a/src/main/scala/exu/issue_slot.scala
+++ b/src/main/scala/exu/issue_slot.scala
@@ -208,6 +208,10 @@ class IssueSlot(num_slow_wakeup_ports: Int)(implicit p: Parameters)
       }
    }
 
+   assert (!(io.ldspec_dst.valid && io.ldspec_dst.bits === 0.U),
+      "Loads to x0 should never speculatively wakeup other instructions")
+
+
    // TODO disable if FP IQ.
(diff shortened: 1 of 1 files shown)


src/main/scala/exu/issue_slot.scala
```diff
@@ -224,10 +228,12 @@ class IssueSlot(num_slow_wakeup_ports: Int)(implicit p: Parameters)
 
    when (io.ldspec_miss && slot_p1_poisoned)
    {
+      assert(slotUop.pop1 =/= 0.U, "Poison bit can't be set for pop1=x0!")
       updated_p1 := false.B
    }
    when (io.ldspec_miss && slot_p2_poisoned)
    {
+      assert(slotUop.pop2 =/= 0.U, "Poison bit can't be set for pop2=x0!")
```
ccelio, line 236: Sorry I'm not looking at the code, but what happens if pop1/pop2 are DontCares?

jerryz123: `slot_pX_poisoned` shouldn't be set in that case, and the previous conditions maintain that.

ccelio (approved):

jerryz123:
I'll merge this when I confirm the additional asserts don't fire when booting Fedora

PR #166, opened 2019-01-24
Support RV32 configurations

### LINKED ISSUE #164: Allow RV32 BOOM Configs
- [x] Support RV32I
- [ ] Support RV32IF

jerryz123:
Resolves #164


src/main/scala/common/consts.scala
```diff
@@ -23,7 +23,7 @@ import freechips.rocketchip.util.Str
 trait BOOMDebugConstants
 {
    val DEBUG_PRINTF        = false // use the Chisel printf functionality
-   val COMMIT_LOG_PRINTF   = false // dump commit state, for comparision against ISA sim
+   val COMMIT_LOG_PRINTF   = true // dump commit state, for comparision against ISA sim
```
abejgonzalez: Keep disabled for merge. Forgot to undo?

src/main/scala/common/parameters.scala
```diff
@@ -127,6 +126,8 @@ trait HasBoomCoreParameters extends freechips.rocketchip.tile.HasCoreParameters
    val usingFDivSqrt = boomParams.fpu.isDefined && boomParams.fpu.get.divSqrt
 
    val mulDivParams = boomParams.mulDiv.getOrElse(MulDivParams())
+   // TODO: Allow this
```
abejgonzalez: Extraneous comment since it now "works"

src/main/scala/system/Configs.scala
```diff
@@ -27,8 +27,11 @@ class BoomConfig extends Config(
    new DefaultBoomConfig ++
    new WithNBoomCores(1) ++
    new WithoutTLMonitors ++
-   new freechips.rocketchip.system.BaseConfig)
+      new freechips.rocketchip.system.BaseConfig)
```
abejgonzalez: Spacing?

src/main/scala/system/Configs.scala
```diff
@@ -119,6 +124,17 @@ class TracedSmallBoomConfig extends Config(
    new WithoutTLMonitors ++
    new freechips.rocketchip.system.BaseConfig)
 
+//RV32IMAC TODO: Support FP
```
abejgonzalez, line 127: I feel like this should be mentioned in the future work section of the docs or at the very least in a GH issue

abejgonzalez:
Lgtm other than the failing csmith tests which I am looking into. Ill approve once that is figured out.

abejgonzalez (approved):
Lgtm. CI was failing since the emulator was taking ~16m on that seed. So I increased the timeout and reduced the number of runs to reduce the overall runtime. I think that expanding the timeout further would be unwarranted (just in case we hit a "looping" case where the sim is not in a pipeline hang but is making no progress)

Just change the recommendations then I think this is mergable

src/main/scala/system/Configs.scala
jerryz123: I'll add to issue

PR #171, opened 2019-01-27
[exu] Deprecate unpipelined integer multiplier

jerryz123:
This will also simplify future changes to execute pipe

### DIFF
--- a/src/main/scala/exu/execute.scala
+++ b/src/main/scala/exu/execute.scala
@@ -293,12 +293,11 @@ class ALUExeUnit(
    }
 
    // Div/Rem Unit -----------------------
-   // Note: We use the PipelinedMul unit for muls
-   var div: MulDivUnit = null
+   var div: DivUnit = null
    val div_resp_val = WireInit(false.B)
    if (has_div)
    {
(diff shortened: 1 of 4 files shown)


src/main/scala/exu/execute.scala
```diff
@@ -297,30 +292,29 @@ class ALUExeUnit(
       fu_units += fdivsqrt
    }
 
-   // Mul/Div/Rem Unit -----------------------
-   var muldiv: MulDivUnit = null
-   val muldiv_resp_val = WireInit(false.B)
-   if (has_muldiv)
+   // Div/Rem Unit -----------------------
+   // Note: We use the PipelinedMul unit for muls
+   var div: MulDivUnit = null
```
abejgonzalez: I think it would be worth it to rename this to DivUnit ... then in the class just specify that we use the MulDiv from rocket to just do div.

abejgonzalez: Also you did the majority of the work renaming things, it would make sense to change this class also.

abejgonzalez (approved):

PR #172, opened 2019-01-29
Rocket jan2019

pentin-as:
Integrated with rocket-chip from 01/12/2019

The boom files duplicated in the rocket-chip have been updated. If the files did not change relative to the rocket-chip, the rocket-chip edition was accepted; if they did, the boom edition was changed.

No corrections were applied for the implementation of clock gating.
No corrections were applied for the implementation of instructions: CEASE, CFLUSH_D_L1, C_NOP, xxx_Q, FCVT_Q_xxx, FMV_Q_X.

This requires new scripts in the boom-template (because riscv-tools is no longer included in the rocket-chip).


abejgonzalez:
As an FYI, we are waiting to look at this PR and the related one in boom-template until rocket-chip has been merged/bumped in FireSim. This will make it easier for external users using that FireSim to use BOOM also. So we didn't forget both PRs!

pentin-as:
Error with two cash misses:
Assertion failed
     at Broadcast.scala: 261 assert (! sent_d)

Request from DCache in two addresses: 0x10023c00 and 0x10002800.
screenshot from 2019-02-11 18-07-43

Good exchange:
screenshot from 2019-02-11 18-10-20

Broadcast forms the wrong diagram and because of this assert.
screenshot from 2019-02-11 18-11-56

This error has already been fixed here:
https://github.com/freechipsproject/rocket-chip/pull/1690

ccelio:
Thanks @pentin-as !

.circleci/create-hash.sh
```diff
@@ -31,6 +31,7 @@ elif [ $1 == "rocket-chip" ]; then
     cd rocket-chip
     git fetch
     git checkout $(cat ../boom/ROCKETCHIP_VERSION)
+    git submodule update --recursive
```
abejgonzalez, line 34: Here you just want to create the hash. Therefore, you don't need to init the submodules yet (this is done for you when you run the `prepare-for-rtl-build.sh`. Is there a specific reason why you did this?

pentin-as: I had a problem - the prepare-smallboomconfig test was falling.
https://circleci.com/gh/riscv-boom/riscv-boom/409?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
The reason for the fall is chisel and firrtl old, and the rocketchip is new.
I see in this task only two scripts call: .circleci/create-hash.sh and .circleci/do-rtl-build.sh. git checkout rocketchip is in the first of them, that's where I added the update chisel and firrtl.

Actually, it's a mystery to me how this can work :(
I don't know exactly how circleci works. I assume that each task is performed independently of the other, from a clean slate (except for the caches mechanism), possibly on different servers. But here, git submodule update --init is not explicitly called anywhere, after cloning the boom-template, besides the .circleci/prepare-for-rtl-build.sh file, which is called in the prepare-build-environment task. Of course, the prepare-smallboomconfig task requires the prepare-build-environment task, but it is still another task.

abejgonzalez: I agree, CircleCI is a bit weird, the way that I set it up is as the following...

`install-riscv-tools` -> `prepare-build-environment` -> `prepare-*config` -> `*-tests`

Each stage is run with a clean container (except for the cache as you mentioned). The `riscv-tools` version is cached using the `rocket-chip.hash` (in the `install-riscv-tools` step). And the `boom-template` is also cached in the `prepare-build-environment` step. I think this issue is related to `boom-template` master being used for CircleCI. Thus, the hash of `boom-template` didn't change leading to the cache to not detect an update. I would need to see a link to a full workload to see for sure

abejgonzalez:
In general, I think that this PR looks good. 

A question that we can answer later is... do we want to create our own "OMBoomCore" classes/model? I don't exactly know what these are used for yet but they are not integral to getting the verilog to output (so far as I know). 

As for the CircleCI changes... Berkeley Architecture Research is looking to depreciate `boom-template` in favor of a top-level repo that has all of the Berkeley research projects (rocket-chip, hwacha, boom, etc). So most of the CI stuff will have to change. For now, I think this is good but changes will happen.

Finally can you rebase/merge these changes with the current master branch and have the tests run. If that passes Ill go ahead and merge things in.

Other than those things, I think this is about ready to merge. 

Once again thanks so much for the PR!

abejgonzalez (approved):

PR #173, opened 2019-01-29
[rvc] Fix RVC compatibility with banked I$

jerryz123:
Some fetch bundles only receive half their bits due to banking.

### DIFF
--- a/src/main/scala/ifu/fetch.scala
+++ b/src/main/scala/ifu/fetch.scala
@@ -244,9 +244,12 @@ class FetchControlUnit(fetch_width: Int)(implicit p: Parameters) extends BoomMod
    val prev_half    = Reg(UInt(coreInstBits.W))
    // Tracks if last fetchpacket contained a half-inst
    val prev_is_half = RegInit(false.B)
-   // Tracks previously fetched pc
-   val prev_fetchedpc = Reg(UInt(vaddrBitsExtended.W))
-   val use_prev = prev_is_half && f3_fetch_bundle.pc === prev_fetchedpc + fetchBytes.U
+   // Tracks nextpc after the previous fetch bundl
+   val prev_nextpc = Reg(UInt(vaddrBitsExtended.W))
+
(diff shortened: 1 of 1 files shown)


src/main/scala/ifu/fetch.scala
```diff
@@ -244,9 +244,12 @@ class FetchControlUnit(fetch_width: Int)(implicit p: Parameters) extends BoomMod
    val prev_half    = Reg(UInt(coreInstBits.W))
    // Tracks if last fetchpacket contained a half-inst
    val prev_is_half = RegInit(false.B)
-   // Tracks previously fetched pc
-   val prev_fetchedpc = Reg(UInt(vaddrBitsExtended.W))
-   val use_prev = prev_is_half && f3_fetch_bundle.pc === prev_fetchedpc + fetchBytes.U
+   // Tracks nextpc after the previous fetch bundl
```
abejgonzalez: *bundle

src/main/scala/ifu/fetch.scala
```diff
@@ -270,6 +273,15 @@ class FetchControlUnit(fetch_width: Int)(implicit p: Parameters) extends BoomMod
          // Need special case since 0th instruction may carry over the wrap around
          inst     := f3_imemresp.data(i*coreInstBits+2*coreInstBits-1,i*coreInstBits)
          is_valid := use_prev || !(f3_valid_mask(i-1) && f3_fetch_bundle.insts(i-1)(1,0) === 3.U)
+      } else if (fetchBytes > 8 && i == fetch_width / 2 - 1) {
```
abejgonzalez: Dos fetchBytes > * determine if it is banked or not? If so, is there a better way to determine if the cache is banked rather than this hardcoded value.

abejgonzalez: Also, I would use a () for (fetch_width/2)-1 when possible.

abejgonzalez (approved):

PR #175, opened 2019-01-29
[CI] Re-enable BoomConfig and change csmith cmd

abejgonzalez:


jerryz123:
BoomConfig seems to build. MegaBoomConfig seems to still fail. 

Can we change is into a real PR that enables the BoomConfig regressions?

abejgonzalez:
Okay, I rebased to squash the commit into a cleaner one and re-pushed. Should be good to merge... assuming CI passes.

jerryz123 (approved):

PR #176, opened 2019-01-29
Refactor the execution unit generation

jerryz123:
The previous code had a lot of hard hacks to map execution units
to read ports, write ports, issue units, etc.

These changes hope to reduce the number of hard hacks and clarify
the core. This will also make it easier to generate a unified-issue
machine.

Sorry about the massive commit, this couldn't really be done incrementally.

In summary, the changes are
- Explicitly label destinations of exe unit responses, whether they get a port to IRF or FRF (iresp/fresp), or if they share a long-latency port (ll_iresp/ll_fresp)
- Cleanup usage of attributes carried by the ExecutionUnits collection
- Move IFPU to the integer pipeline execution unit collection.

### DIFF
--- a/src/main/scala/exu/core.scala
+++ b/src/main/scala/exu/core.scala
@@ -74,46 +74,53 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
 
    // Only holds integer-registerfile execution units.
    val exe_units = new boom.exu.ExecutionUnits(fpu=false)
+
    // Meanwhile, the FP pipeline holds the FP issue window, FP regfile, and FP arithmetic units.
    var fp_pipeline: FpPipeline = null
-   if (usingFPU) fp_pipeline = Module(new FpPipeline())
+   if (usingFPU) fp_pipeline = Module(new FpPipeline)
 
    // ********************************************************
    // Clear fp_pipeline before use
    if (usingFPU){
-     fp_pipeline.io.ll_wport := DontCare
+     fp_pipeline.io.ll_wport  := DontCare
      fp_pipeline.io.wb_valids := DontCare
-     fp_pipeline.io.wb_pdsts := DontCare
+     fp_pipeline.io.wb_pdsts  := DontCare
    }
 
-   val num_fp_wakeup_ports = if (usingFPU) fp_pipeline.io.wakeups.length else 0
 
-   val num_irf_write_ports = exe_units.map(_.num_rf_write_ports).sum
-   val num_fast_wakeup_ports = exe_units.count(_.isBypassable)
-   val num_wakeup_ports = num_irf_write_ports + num_fast_wakeup_ports
+
+   val num_irf_write_ports    = exe_units.num_irf_write_ports
+   val num_ll_irf_write_ports = exe_units.num_ll_irf_write_ports
+   val num_irf_read_ports     = exe_units.num_irf_read_ports
+
+   val num_fast_wakeup_ports  = exe_units.count(_.bypassable)
+
+   val num_int_wakeup_ports   = num_irf_write_ports + 1 + num_fast_wakeup_ports // + 1 for ll_wb
+   val num_fp_wakeup_ports    = if (usingFPU) fp_pipeline.io.wakeups.length else 0
+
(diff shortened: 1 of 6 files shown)


src/main/scala/exu/fppipeline.scala
```diff
@@ -62,35 +62,28 @@ class FpPipeline(implicit p: Parameters) extends BoomModule()(p) with tile.HasFP
...
-                                 exe_units.withFilter(_.uses_iss_unit).map(e => e.num_rf_read_ports).sum,
-                                 // TODO get rid of -1, as that's a write-port going to IRF
-                                 exe_units.withFilter(_.uses_iss_unit).map(e => e.num_rf_write_ports).sum - 1 +
-                                    num_ll_ports,
+                                 exe_units.num_frf_read_ports,
+                                 exe_units.num_frf_write_ports + 1, // + 1 for ll writeback
                                  fLen+1,
-                                 exe_units.bypassable_write_port_mask
+                                 // No bypassing for FP
+                                 Seq(false) ++ (0 until exe_units.num_frf_write_ports).map(x=>false)
```
abejgonzalez: How about being more concise?

Seq.fill(exe_units.num_frf_write_ports + 1){ x => false }

abejgonzalez: I think that is easier to read than what was previously there

src/main/scala/exu/execution_units.scala
```diff
@@ -167,45 +164,39 @@ class ExecutionUnits(fpu: Boolean = false)(implicit val p: Parameters) extends H
...
-         mask
-      }
-      else
-      {
-         // The mem-unit will also bypass writes to readers in the RRD stage.
-         exe_units.withFilter(_.usesIRF).map(u => u.isBypassable || u.is_mem_unit)
-      }
+   if (!fpu)
+   {
+      // if this is for FPU units, we don't need a memory unit (or other integer units)..
```
abejgonzalez: Double period!

src/main/scala/exu/fppipeline.scala
```diff
@@ -62,35 +62,28 @@ class FpPipeline(implicit p: Parameters) extends BoomModule()(p) with tile.HasFP
...
                                  ))
    val fregister_read   = Module(new RegisterRead(
                            issue_unit.issue_width,
-                           exe_units.withFilter(_.uses_iss_unit).map(_.supportedFuncUnits),
-                           exe_units.withFilter(_.uses_iss_unit).map(_.num_rf_read_ports).sum,
-                           exe_units.withFilter(_.uses_iss_unit).map(_.num_rf_read_ports),
-                           exe_units.num_total_bypass_ports,
+                           exe_units.withFilter(_.reads_frf).map(_.supportedFuncUnits),
+                           exe_units.num_frf_read_ports,
+                           exe_units.withFilter(_.reads_frf).map(x => 3),
```
abejgonzalez, line 75: There are two places where the 3 is copied to indicate the amount of FP rd ports. Can this put as a constant somewhere? Even though it is unlikely that there will be more than 3 read ports in the future I think this would be nice QoL improvement.

src/main/scala/exu/core.scala
```diff
@@ -144,15 +151,22 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
 
    // Branch Unit
    val br_unit = Wire(new BranchUnitResp())
    val brunit_idx = exe_units.br_unit_idx
    br_unit <> exe_units.br_unit_io
 
+   for (eu <- exe_units) {
+      eu.io.brinfo := br_unit.brinfo
+      eu.io.com_exception := rob.io.flush.valid
+   }
```
abejgonzalez, line 167: Spacing.

src/main/scala/exu/fppipeline.scala
jerryz123: I think its clearer to hard-set it here, since it will be more immediately obvious that this is just stating that FP instructions might have 3 operands.

abejgonzalez: This is fine for now then.

src/main/scala/exu/core.scala
jerryz123: What do you mean here?

abejgonzalez (approved):

PR #180, opened 2019-02-06
[ftq] FTQ should prioritize flush over branch mispredict

jerryz123:


ccelio:
I recommend a slightly different way of fixing this. There is a stylistic error being performed in the original code.... as you've noticed, there are two separate `when` blocks driving the `enq_ptr.value` signal. But instead of snaking the previous when's firing condition to gate off the later when, I recommend you organize the code slightly differently so that there is only a single when/elsewhen block that drives `enq_ptr.value` , with the `when` for the flush being higher priority. This will point you to a second mistake that was missed... the `maybe_full` signal also needs to be corrected when a flush occurs.

ccelio (changes requested):
combine whens, fix maybe_full too.

ccelio (approved):

PR #181, opened 2019-02-07
[all] Cleanup code base

abejgonzalez:
Things done:
- [x] Cleanup headers, spacing
- [x] Added `IO()` to missing bundles
- [x] Newline for `{` (for consistent code brackets)
- [x] renamed `ifu/fetch` to `ifu/fetch-control-unit`
- [x] ScalaDocs for all classes
- [x] Add more comments and spaces where necessary
- [x] Which syntax for multi-line arguments to arg, instead of ,arg
- [ ] Rename "unclear" variables (subjective but this is low priority)
- [x] Logical folder hierarchy for subsections of core
- [x] Rename files to be consistent (use word-word format)


abejgonzalez:
@jerryz123 @ccelio This is the PR that I am using to track code cleanup. As you can see, I ended up renaming the files to a "word-word" format. I also ended up reorganizing the BPU a bit to be more accessible (know what files correspond with what files). Let me know what you guys think.

src/main/scala/bpu/bpd/br-predictor.scala
```diff
@@ -321,53 +320,3 @@ object BrPredictor
...
-//------------------------------------------------------------------------------
-
-// Act as a "null" branch predictor (it makes no predictions).
-class NullBrPredictor(
-   fetch_width: Int,
-   history_length: Int = 12
-   )(implicit p: Parameters) extends BrPredictor(fetch_width, history_length)(p)
-{
-   override def toString: String = "  Building (0 kB) Null Predictor (never predict)."
-   io.resp.valid := false.B
```
jerryz123, line 336: Are these broken right now? I think these null predictors are useful to keep around as a point of comparison

jerryz123:
For src code organization: why are execution_units.scala, fu_decode.scala, issue_units.scala, register_read.scala, rename-stage.scala not in their respective folders?

src/main/scala/bpu/bpd/br-predictor.scala
abejgonzalez: I don't know if they are broken. That would be something to check.

abejgonzalez:
Reply, my basic reasoning for this was that I tried to keep the top level module that is called in `core` in the folder with `core`. Im still thinking about this but another option is to move these top level modules into the folders. This is in contrast with `bpu` which does the structure that you refer to. I will probably switch it to a `bpu` like structure.

jerryz123:
> Reply, my basic reasoning for this was that I tried to keep the top level module that is called in `core` in the folder with `core`. Im still thinking about this but another option is to move these top level modules into the folders. This is in contrast with `bpu` which does the structure that you refer to. I will probably switch it to a `bpu` like structure.

I like

```
exu/
   core.scala
   rob.scala
   decode.scala
   fp-pipeline.scala
   execution-units/
   issue-units/
   regfiles/
   rename/
```

abejgonzalez:
Ok @jerryz123 @ccelio, I think this is the final version of the reorganization + file renaming. Let me know if you all have comments on it. After I get feedback, Ill fix the conflict(s) and merge the PR.

.gitignore
```diff
@@ -1,3 +1,4 @@
+*.swp
```
jerryz123, line 1: Out of curiosity, what's generating .swp files?

abejgonzalez: My vim :)

abejgonzalez: I figured for all swap file users this might be nice.

jerryz123:
LGTM, can we run the CI to confirm everything builds?

jerryz123 (approved):

abejgonzalez:
Yup. I will run it soon. Just need to merge in updated changes.

PR #182, opened 2019-02-08
[ftq] FTQ should update internal commit_ptr on flushes

jerryz123:
This is to resolve a pipeline hang when the core fetches no valid instructions. In this case, the FTQ receives repeated `flush` signals, but no `deq` signals, causing it to fill and stall the frontend.

I believe the issue is that excepting instructions don't send the ftq_idx's to the FTQ to be dequeued. In normal program execution, this doesn't break the FTQ since a later valid committed instruction would increment the commit_ptr past the entry of the excepting instruction. In the case where the core is only fetching invalid instructions (such as when the bootrom is invalid). the core should be able to spin on invalid instruction exceptions.

I'm not convinced that the flush ftq_idx and the commit ftq_idx will always match, however.

### DIFF
--- a/src/main/scala/common/consts.scala
+++ b/src/main/scala/common/consts.scala
@@ -259,8 +259,9 @@ trait ScalarOpConstants
    val uopFSQRT_S   = 103.U(UOPC_SZ.W)
    val uopFSQRT_D   = 104.U(UOPC_SZ.W)
 
-   val uopSYSTEM    = 105.U(UOPC_SZ.W) // pass uop down the CSR pipeline and let it handle it
-   val uopSFENCE    = 106.U(UOPC_SZ.W)
+   val uopWFI       = 105.U(UOPC_SZ.W) // pass uop down the CSR pipeline
+   val uopERET      = 106.U(UOPC_SZ.W) // pass uop down the CSR pipeline, also is ERET
+   val uopSFENCE    = 107.U(UOPC_SZ.W)
 
    // The Bubble Instruction (Machine generated NOP)
    // Insert (XOR x0,x0,x0) which is different from software compiler
--- a/src/main/scala/exu/core.scala
+++ b/src/main/scala/exu/core.scala
@@ -421,15 +421,26 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
 
    // We must redirect the PC the cycle after playing the SFENCE game.
    io.ifu.flush_take_pc     := rob.io.flush.valid || RegNext(lsu.io.exe_resp.bits.sfence.valid)
-   io.ifu.flush_pc          := RegNext(csr.io.evec)
+
+   // TODO FIX THIS HACK
+   // The below code works because of two quirks with the flush mechanism
+   //  1 ) All flush_on_commit instructions are also is_unique,
+   //      In the future, this constraint will be relaxed.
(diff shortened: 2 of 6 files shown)


ccelio (approved):
Good catch and good job adding the assert.  You may even be able to be stricter on the assert -- should a flush and deq.valid ever occur on the same cycle?

jerryz123:
I think you are right, while in the current design flushes may coincide with commits, this should not be the case.

- For eret flushes, I observe `io.flush.valid` and `io.deq.valid` both being high after a eret. The flush signals sent after a eret do not carry the ftq_idx of the eret instruction, but since erets are unique, com_xcpt.ftq_idx points to the right ftq_idx anyways.

- For flush-on-commit instructions, currently all flush-on-commit insns also are marked as unique, this guarantees that the commit bundle contains only the flushing instruction. Since the flush is sent a cycle after the commit, `io.deq.valid` and `io.flush.valid` do not line up.

- For exception flushes, we don't take exceptions unless its at the head of the ROB, so exception-triggered flushes will never coincide with a commit. This is fine.

Is there any reason why `flush_on_commit` and `is_unique` are separate tags for instructions? The only instructions for which the decode tables mark `flush_on_commit != is_unique` are SCALL, SBREAK, MRET, DRET, SRET, and these will throw csr.io.eret anyways, which causes a flush on commit. 
If `flush_on_commit` and `is_unique` have the same meaning, then I can fix this by redoing the flush logic in the ROB to trigger on committing a `is_unique` instruction.

ccelio:
Ah, I forgot `flush.valid` was for more than just exceptions. Then yes, your assert is great. 

Currently, all `flush_on_commit` are also `unique`, but that's just a simplification; for performance reasons, there are some examples where we should `flush_on_commit` but not make them unique.

jerryz123:
I pushed a new commit to fix the eret issues, and also clean up the flush signal generation.

Previously, flush signals from erets were sent too early. This commit fixes and simplifies that logic by decoding eret insts as flush_on_commit, instead of relying on the csr to report that.

This should also make future optimizations in splitting up is_unique and flush_on_commit easier.

jerryz123:
@ccelio Any issues with making flushes consistently 1-cycle behind commits? 

Why are flushes 1-cycle late anyways? I see there are some comments indicating flushes should be timed together with the commits and exceptions.

ccelio:
This commit is all I can quickly find that reveals my intentions: https://github.com/riscv-boom/riscv-boom/commit/08767164ecbd446f89a1fb0f0d4f1490c0795639

> [rob] Refactor xcpt IOs, flush timings.
>   - The ROB consults the CSRFile on cycle 0, and exception signals are
>     communicated to the CSRFile on cycle 0.
>   - The pipeline flush signals (and PC redirect) are now performed on cycle 1.

That suggests that I am a slave to whatever the CSRFile is trying to communicate to the pipeline (since it accepts exception signals and then tells us where to go). 

Now in this case, I'm not really distinguishing between "flushes because I feel like it" and "flush the pipeline because we encountered an exception and we need guidance from the CSRFile on what to do".

More things to think about: flushes should be rare, so another cycle is "whatever". Also, the flush signal gets broadcast to a ton of places, so it's nice to pipeline it and let register retiming help you out (it will be a very slow signal).

That's some of my rationale. I have no opinions on your changes, but feel free to go with what seems correct to you.

Issue #178, opened 2019-02-01
Recovery after executing 'x


pentin-as:
MegaBoomConfig

Assertion failed: Pipeline has hung.
    at core.scala:1143 assert (!(idle_cycles.value(13)), "Pipeline has hung.")

This happens as the processor goes to execution from the address 0x10022f88, in which the 'x.

I have the code:
    10001b5c:	64813083          	ld	ra,1608(sp)
    10001b60:	65010113          	addi	sp,sp,1616
    10001b64:	00008067          	ret

sp = 100222f0
When executed, it becomes ra = sp + 1608 + 1616 = 0x10022f88.

This happens because a false bypass happens in iregister_read:
In iregister_read, I see the following:
![screenshot from 2019-02-01 17-01-43](https://user-images.githubusercontent.com/30556064/52121940-265ffd00-2643-11e9-863c-ce3dc2a8c287.png)

This bypass goes from alu to csr_exe_unit:
![screenshot from 2019-02-01 17-06-09](https://user-images.githubusercontent.com/30556064/52122131-b736d880-2643-11e9-95df-99cd32e0f05d.png)

It seems to me that this bypass should not be.
I will continue to study the problem. Does anyone have any ideas?

ccelio:
Is the test small and reproducible using the default Boom flow? If so, maybe we can take a look.

However, if you have a fsdb file of the failure occurring, then you have everything needed to diagnosis this I think.

Regarding the bypass logic, the only thing that would prevent the `add` from the future bypassing to the `load` behind him is that the `add` should be allocated a new, different physical register specifier (`pdst`). I would check to see what the `pdst` of the bypassed `add` is and compare it to the `pop1` (operand) of the `load`. I assume they will be the same, if this erroneous bypass is occurring.

Then the next step is to figure out who has the wrong physical specifier and why. Did they enter the issue window with the wrong specifier? Did they leave rename with the wrong specifier? Was there a stall in rename? Did the load and add leave rename together? Was there a mispredict and reset?

pentin-as:
The problem was that I simulated without RANDOMIZE_MEM_INIT. Because of the false predictions, boom switched to executing the 'x instruction, which led to the breakdown of FetchController and "Pipeline has hung."

Here is a small test that reproduces this situation:
```
  li   a0, 7
wait:
  addi a0, a0, -1
  bne  a0, x0, wait
  beq  a0, x0, after_wait
kill_boom:
[...]

PR #183, opened 2019-02-08
Fixed a false value in history when flush happened

pentin-as:
screenshot from 2019-02-08 12-52-25
screenshot from 2019-02-08 12-56-08

### DIFF
--- a/src/main/scala/ifu/fetchtargetqueue.scala
+++ b/src/main/scala/ifu/fetchtargetqueue.scala
@@ -270,7 +270,7 @@ class FetchTargetQueue(num_entries: Int)(implicit p: Parameters) extends BoomMod
 
    when (io.restore_history.valid)
    {
-      val ridx = Mux(io.flush.valid, io.com_ftq_idx, io.brinfo.ftq_idx)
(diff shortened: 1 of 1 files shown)


jerryz123:
Nice catch, it seems like before, all flushes were messing up the branch history?

Also, how are you running simulations with X values? When I disable random-mem-init, x-propagation blows up the simulation.

pentin-as:
I'm trying to make it possible to work without RANDOMIZE_MEM_INIT here:
https://github.com/riscv-boom/riscv-boom/issues/178
This is not finished yet.
I have 'x because I work without RANDOMIZE_REG_INIT. To do this, I call (and after compilation, cancel) the following patch:
[fix_x_pessimism_boom.patch.txt](https://github.com/riscv-boom/riscv-boom/files/2855322/fix_x_pessimism_boom.patch.txt)
I tried to create an issue to initialize RRArbiter, but the chisel developers disagree:
https://github.com/freechipsproject/chisel3/pull/724
RRArbiter is used only in PTW. But doing fork PTW with my own RRArbiterInit seemed like a bad idea to me, so I do used patches.

ccelio (approved):

PR #187, opened 2019-02-13
[config] Update SmallBoomConfig Cache Size

abejgonzalez:
SmallBoomConfig previously inherited the BoomConfig cache sizes. Instead, have it match the MediumBoomConfig sizes.

@ccelio  Just to make sure was this missed or was this done for a reason?

### DIFF
--- a/src/main/scala/common/configs.scala
+++ b/src/main/scala/common/configs.scala
@@ -117,7 +117,8 @@ class WithSmallBooms extends Config((site, here, up) => {
          tage = Some(TageParameters(enabled=false)),
          bpdBaseOnly = Some(BaseOnlyParameters(enabled=true)),
          nPerfCounters = 2),
-      icache = Some(r.icache.get.copy(fetchBytes=2*4))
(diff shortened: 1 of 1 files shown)


ccelio:
The config overriding is pretty easy to screw up and make mistakes on, so there can easily be "small" configurations that end up with default values that turn them into fairly large designs.

The original configurations were set up to be the following:

* SmallBoomConfig -- allow for the fastest compile to test out new features.
* MediumBoomConfig -- match the BROOM tape-out configuration (also roughly a Cortex A9).
* MegaBoomconfig -- be the max reasonable configuration designed to stress Chisel/firrtl, but perhaps not have good area efficiency or clock frequency.
* DefaultBoomConfig -- a reasonable configuration to use, experiment on, and tape-out. 

The only really important thing here I think is to realize that new users who download and evaluate BOOM will be using the default CONFIG, so it should be a configuration we want people to use and report performance numbers with.

ccelio (approved):

abejgonzalez:
Yup. I just wanted to make sure that there was nothing I missed in particular. Thanks @ccelio

PR #188, opened 2019-02-13
[config] configuration updates

abejgonzalez:
Updated SmallBoomConfig to have GShare branch predictor.

Any more recommendations? @donggyukim @jerryz123 @bkorpan 

Also, let's hold off on merging this until the CI commit (in boom-template) is put in (so we can double check this before merging into master)

### DIFF
--- a/src/main/scala/common/configs.scala
+++ b/src/main/scala/common/configs.scala
@@ -113,8 +113,7 @@ class WithSmallBooms extends Config((site, here, up) => {
          numFpPhysRegisters = 48,
          numLsuEntries = 8,
          maxBrCount = 4,
-         tage = Some(TageParameters(enabled=false)),
-         bpdBaseOnly = Some(BaseOnlyParameters(enabled=true)),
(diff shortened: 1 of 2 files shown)


src/main/scala/common/configs.scala
```diff
@@ -113,8 +113,7 @@ class WithSmallBooms extends Config((site, here, up) => {
          numFpPhysRegisters = 48,
          numLsuEntries = 8,
          maxBrCount = 4,
-         tage = Some(TageParameters(enabled=false)),
-         bpdBaseOnly = Some(BaseOnlyParameters(enabled=true)),
+         gshare = Some(GShareParameters(enabled=true, history_length=11, num_sets=2048)),
```
abejgonzalez, line 138: Note, I made the set# = 2^history_length so that the GHR doesn't need to be folded over (for reference look at MediumBoomConfig).

jerryz123 (changes requested):
Can you address #189 in this PR as well?

jerryz123 (approved):

PR #193, opened 2019-02-16
[pnr] Initial simple point-of-no-return implementation in ROB

jerryz123:
This is taken from work on the secureboom project.
A simple PNR head that advances one ROB row per cycle.
Currently drives nothing.

The LSU should eventually be able to mark loads as safe to eventually commit, when they are guaranteed to not cause a memory-ordering failure. This will substantially improve the run-ahead of the PNR. I left a signal from the LSU to control this, but kept it tied off for now.

### DIFF
--- a/src/main/scala/exu/rob.scala
+++ b/src/main/scala/exu/rob.scala
@@ -192,7 +192,7 @@ class DebugRobSignals(implicit p: Parameters) extends BoomBundle()(p)
 {
    val state = UInt()
    val rob_head = UInt(ROB_ADDR_SZ.W)
-   val pnr_head = UInt(ROB_ADDR_SZ.W)
+   val rob_pnr = UInt(ROB_ADDR_SZ.W)
    val xcpt_val = Bool()
    val xcpt_uop = new MicroOp()
    val xcpt_badvaddr = UInt(xLen.W)
@@ -228,8 +228,8 @@ class Rob(
    //commit entries at the head, and unwind exceptions from the tail
    val rob_head     = RegInit(0.U(log2Ceil(num_rob_rows).W))
    val rob_tail     = RegInit(0.U(log2Ceil(num_rob_rows).W))
-   val pnr_head     = RegInit(0.U(log2Ceil(num_rob_rows).W))
-   chisel3.experimental.dontTouch(pnr_head)
+   val rob_pnr     = RegInit(0.U(log2Ceil(num_rob_rows).W))
+   chisel3.experimental.dontTouch(rob_pnr)
    val rob_tail_idx = rob_tail << log2Ceil(width).U
 
    val will_commit         = Wire(Vec(width, Bool()))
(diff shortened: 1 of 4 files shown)


jerryz123:
@bkorpan If there are no concerns here I'd like to merge this.

src/main/scala/exu/rob.scala
```diff
@@ -190,6 +192,7 @@ class DebugRobSignals(implicit p: Parameters) extends BoomBundle()(p)
 {
    val state = UInt()
    val rob_head = UInt(ROB_ADDR_SZ.W)
+   val pnr_head = UInt(ROB_ADDR_SZ.W)
```
ccelio: I'm sorry this is nit-picky, but for consistency, should this be `rob_head`, `rob_pnr`, and `rob_tail`?

src/main/scala/exu/rob.scala
```diff
@@ -704,6 +721,20 @@ class Rob(
    {
       rob_head := WrapInc(rob_head, num_rob_rows)
    }
+   assert(!(rob_head === WrapInc(pnr_head, num_rob_rows) && pnr_head =/= rob_tail),
```
ccelio: Good assert. I'm wondering if there's another assert we can add to catch the scenario where we have an exception thrown on an entry marked "safe". We want to catch the scenario where the pnr erroneously advanced too far.

src/main/scala/lsu/lsu.scala
```diff
@@ -164,6 +166,10 @@ class LoadStoreUnit(pl_width: Int)(implicit p: Parameters,
 {
    val io = IO(new LoadStoreUnitIO(pl_width))
 
+   // TODO: This needs to be high when we find a load that will not fail a mem-ordering check
+   io.ld_success := false.B
+   io.ld_success_rob_idx := 0.U
```
ccelio: Are these signals set to true anywhere? It doesn't look to me that the pnr_safe is being set for loads/stores as early as they can be.

ccelio (changes requested):

src/main/scala/lsu/lsu.scala
ccelio: I see, your commit message explains this.

ccelio:
> The LSU should eventually be able to mark loads as safe to eventually commit, when they are guaranteed to not cause a memory-ordering failure. This will substantially improve the run-ahead of the PNR. I left a signal from the LSU to control this, but kept it tied off for now.

I don't think that's an issue. Loads are "safe" once their address is known. Sure... they might actually throw an ordering-violation exception, but in that scenario, an older load or store will be marked "unsafe" as its address isn't known yet. Once the older address is known, then the younger load notifies the ROB it will throw a "mini-exception" and it can clear its `rob_safe` at the same time. The only trick here is to verify the older load doesn't get marked unbusy and safe (so the ROB head and PNR can proceed) before the younger load can be marked "unsafe" (so the PNR can proceed).

ccelio (approved):
Thanks for addressing my feedback; this looks good to me.

PR #194, opened 2019-02-17
[ras] Work towards re-enabling the RAS

jerryz123:
This RAS is technically re-enabled, but performance is terrible. It is probably not working correctly. There are several points which confuse me.
 - Why do we only take the RAS value if we hit in the BTB arrays? 
 - When F3 sees a JAL that was missed by the BTB, the next fetchpacket is not ignored. In the following code segment, where each instruction is on a different fetchpacket,
```
jal <funct1>
jal <funct2>
```
The branch-checker will send updates to the RAS for both packets, messing up the RAS stack. This makes me doubt if RAS pop operations should be done in F3, in BranchDecode. 

Branch prediction is hard.

### DIFF
--- a/src/main/scala/bpu/bpd-pipeline.scala
+++ b/src/main/scala/bpu/bpd-pipeline.scala
@@ -143,20 +143,19 @@ class BranchPredictionStage(fetch_width: Int)(implicit p: Parameters) extends Bo
 
    //************************************************
    // Update the RAS
-   // TODO XXX  reenable RAS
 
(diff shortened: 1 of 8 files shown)


jerryz123:
Closing this because need to approach this more incrementally

ccelio:
You want the RAS to make predictions quickly -- before you decode the actual instructions from the I$. Therefore, you have to rely on the BTB to provide the instruction type prediction.

PR #195, opened 2019-02-17
[ifu] Reduce the number of RVCExpanders that get instantiated

jerryz123:


src/main/scala/ifu/branch-checker.scala
```diff
@@ -50,6 +50,7 @@ class BranchChecker(fetch_width: Int)(implicit p: Parameters) extends BoomModule
       val is_jal        = Input(Vec(fetch_width, Bool()))
       val is_jr         = Input(Vec(fetch_width, Bool()))
       val is_call       = Input(Vec(fetch_width, Bool()))
+      val is_ret        = Input(Vec(fetch_width, Bool()))
```
abejgonzalez, line 53: This is not used in the module... is this just being used for consistency in having all the decoded signals enter this module?

abejgonzalez: This applies also the the `is-ret` wire that you added to the ftcontroller.

jerryz123: This signal will likely be useful when re-enabling the RAS.

src/main/scala/exu/decode.scala
```diff
@@ -573,6 +577,12 @@ class BranchDecode(implicit p: Parameters) extends Module
    io.is_br   := cs_is_br
    io.is_jal  := cs_is_jal
    io.is_jalr := cs_is_jalr
+   io.is_call := (cs_is_jal || cs_is_jalr) && GetRd(inst) === RA
+   io.is_ret  := cs_is_jalr && GetRs1(inst) === BitPat("b00?01")
+
+   val b_imm32 = Cat(Fill(20,inst(31)),inst(7),inst(30,25),inst(11,8),0.U(1.W))
```
abejgonzalez: Instead of doing this copying form the Compute***Target functions... Why not just make the e_inst be an input to the helper function and specify that the function requires an expanded instruction (this will move the RVCExpander outside of the function). This would help if you want to reuse the instruction.

jerryz123: Compute***Target is used elsewhere with uncompressed instructions.

abejgonzalez (approved):

PR #196, opened 2019-02-18
[ren] Allow dynamic assignment of rename busy table read ports

jerryz123:
This PR allows us to generate fewer busy table read ports for FP registers.


jerryz123:
The dynamic assignment logic can be moved into rename-1 if necessary

PR #198, opened 2019-02-20
Do not connect the PTW if not usingPTW

jerryz123:


src/main/scala/lsu/types.scala
```diff
@@ -68,8 +68,7 @@ trait CanHaveBoomPTWModule extends HasBoomHellaCacheModule
 {
   val outer: CanHaveBoomPTW
   val ptwPorts = ListBuffer(outer.dcache.module.io.ptw)
-  val ptw = Module(new PTW(outer.nPTWPorts)(outer.dcache.node.edges.out(0), outer.p))
-  ptw.io <> DontCare // Is overridden below if PTW is connected
+  val ptw = if (outer.usingPTW) Module(new PTW(outer.nPTWPorts)(outer.dcache.node.edges.out(0), outer.p)) else null
```
abejgonzalez: I think this would be a good place to use `Option`.

Aka: `val ptw: Option[PTW] = (outer.usingPTW).option(new PTW(outer.nPTWPorts))`

Then to access it you would just do `ptw.get`. Or (I think) you could do a `ptw.getOrElse(DontCare)` to remove the need for the `usingPTW` check.

abejgonzalez (approved):

abejgonzalez:
Can you confirm that this disconnects the PTW properly? Just want to make sure before the merge.

jerryz123:
Yep PTW does not show up in the generated verilog

PR #199, opened 2019-02-21
Force ICache to generated named memories

jerryz123:


src/main/scala/ifu/icache.scala
```diff
@@ -287,11 +287,12 @@ class ICacheModule(outer: ICache) extends ICacheBaseModule(outer)
     else (nSets * refillCycles)
   val dataArraysB0 = Seq.fill(nWays) { SyncReadMem(ramDepth, UInt(dECC.width(wordBits/nBanks).W)) }
```
abejgonzalez: Currently, it looks like the `if` determines whether to use the `dataArraysB*` or the `dataArrays`. To me it makes more sense that these `SyncReadMem`'s are within the `if` statement blocks. Have you tried doing something like...

`val dataArrays   = Seq.tabulate(nWays) { SyncReadMem(nSets * refillCycles, UInt(dECC.width(wordBits).W)).suggestName("dataArrayWay" + _.toString) }`

abejgonzalez (approved):

PR #201, opened 2019-02-23
[exu] Refactor fp stdatagen as output of FPIU unit

jerryz123:
This is to work towards enabling >1 fp width.

### DIFF
--- a/src/main/scala/exu/execution-units/execution-unit.scala
+++ b/src/main/scala/exu/execution-units/execution-unit.scala
@@ -454,12 +454,13 @@ class FPUExeUnit(
       reads_frf  = true,
       writes_frf = true,
       writes_ll_irf = has_fpiu,
+      writes_irf = has_fpiu, // HACK: the "irf" port actually goes to the LSU fpu SDATAGen
       num_bypass_stages = 0,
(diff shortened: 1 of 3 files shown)


jerryz123:
This only functional side effect this has is that FP store data will be blocked when the FPToInt queue is full. This should be a small effect

ccelio:
I would be most concerned about hurting the timings of FP stores versus INT stores. You want store data to get to the LSU on the same RegRead/Exe(Agen)/Mem pipeline. If you route FP store-data through the Float2Int pipeline, isn't that padded out 4 cycles, and then queued up behind the other F2I write backs? This destroys context switching.

jerryz123:
The fpstdata goes on a different path than fptoint moves. Fpstdata gets to the lsu the cycle after regread.

I'm just repurposing the fptoint bitmask to also apply to fpstdatagen. 

The only way this affects timings is that fpstoredata will not issue while the fptoint queue is full.

ccelio (approved):
Thanks for the clarification; still feel a little nervous wiring `iresp` to `sdq` (as opposed to adding a detected sdq port or sdq "unit"), but it's less hacky than it was before.

PR #208, opened 2019-02-25
[ifu] Add parameter to disable fetch monitor

abejgonzalez:
Useful to disable to save space in EE290C tapeout


jerryz123:
Why is this necessary? The fetch monitor should just be a bunch of asserts, it shouldn't show up in synthesis.

abejgonzalez:
If the tools do a good enough job of trimming the signals/logic out then that is fine... this is not needed then. I guess one thing to check is the module heirarchy that was synthesized and see if it did get rid of all the unneccesary signals. However, I don't think this is a bad thing to specify/parameterize.

jerryz123 (approved):

PR #210, opened 2019-02-28
Instruction buffer in the ROB

jerryz123:
We get a vector of instruction bits at rename-1. We write this into a ROB-sized 1R1W memory, and read it at commit. 

Questions:

- Why is it bad to use io.enq.uop.inst in the ROB for handling ma_fetch? We get this in rename-1, so passing the instruction bits into the ROB seems harmless.
- Getting PCs is trickier, since we store them earlier in the FTQ, and a ROB row may contain uops requesting multiple ftq entries. This needs some additional work to make the trace port reasonable, but is not necessary for RoCC.
- Also is it possible for the FTQ to dequeue entries early? What happens when two uops share the same FTQ entry, and the first is committed before the second? When the first commits, the ROB sends the ftq commit signals. When the second instruction needs to look up its ftq entry, isn't it possible for some other PC to have been written there?

### DIFF
--- a/src/main/scala/common/micro-op.scala
+++ b/src/main/scala/common/micro-op.scala
@@ -21,6 +21,7 @@ import freechips.rocketchip.config.Parameters
 import boom.bpu.BranchPredInfo
 import boom.exu.FUConstants
 
+
 /**
  * Extension to BoomBundle to add a MicroOp
  */
@@ -41,7 +42,7 @@ class MicroOp(implicit p: Parameters) extends BoomBundle()(p)
    val valid            = Bool()
 
    val uopc             = UInt(UOPC_SZ.W)       // micro-op code
-   val inst             = UInt(32.W)
+   val debug_inst       = UInt(32.W)
    val is_rvc           = Bool()
    val pc               = UInt(coreMaxAddrBits.W) // TODO remove -- use FTQ to get PC. Change to debug_pc.
    val iqtype           = UInt(IQT_SZ.W)        // which issue unit do we use?
--- a/src/main/scala/exu/core.scala
+++ b/src/main/scala/exu/core.scala
@@ -1074,6 +1074,7 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
    // Dispatch
    rob.io.enq_valids := rename_stage.io.ren1_mask
    rob.io.enq_uops   := rename_stage.io.ren1_uops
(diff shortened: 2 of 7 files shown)


ccelio:
Questions:

> * Why is it bad to use io.enq.uop.inst in the ROB for handling ma_fetch? We get this in rename-1, so passing the instruction bits into the ROB seems harmless.

Did you mean `enq.uop.pc`?  

>* Getting PCs is trickier, since we store them earlier in the FTQ, and a ROB row may contain uops requesting multiple ftq entries. This needs some additional work to make the trace port reasonable, but is not necessary for RoCC.

The PC is too expensive to pass through the FetchBuffer and into the backend. 

Ignore the trace port. That was designed for single-issue, in-order cores. The RISC-V External Trace Working Group is working on a better proposal that is more amenable to superscalar, OOO cores. Along these lines, I believe the trace port should be sitting off of the FTQ and not the ROB. From the FTQ you can guarantee that all instructions in the commit bundle are aligned to some address boundary and consecutive in addresses. (Of course, for FPGA simulations, the sky is the limit regarding unsynthesizable, inefficient tracing/debug structures.)

Using the `inst` port on the trace port is mostly a non-starter --- your `io.enq.uop.inst` are not actually instructions anymore, thanks to RVC-expansion (and perhaps later, macro-op fusion). So `inst` is a bit of a misnomer, although I can't think of a better name right now. 

>* Also is it possible for the FTQ to dequeue entries early? What happens when two uops share the same FTQ entry, and the first is committed before the second? When the first commits, the ROB sends the ftq commit signals. When the second instruction needs to look up its ftq entry, isn't it possible for some other PC to have been written there?

I'm sure it is possible to use your new PNR to deallocate _many_ things earlier now, including the FTQ entries. 

An FTQ entry is not deallocated until the ROB begins committing instructions that correspond _to the next FTQ entry_.

If you are building an instruction buffer for RoCC instructions, I strongly recommend building a totally separate structure that is orthogonal to the ROB. How often do you expect to  hit RoCC instructions? 128 entries*32 bits is overkill! Your "instruction register file" can be very small, and if it fills up, `decode` stalls.

ccelio (changes requested):
If you still want to push forward on 32b instruction bits for every ROB entry, then I recommend you make it a configuration parameter so it can be turned off.

jerryz123:
I'm happy to punt on the trace port. I was hoping to reuse work between RoCC buffer and the trace port, but if I'm doing just RoCC, I agree a smaller buffer makes more sense.

PR #211, opened 2019-03-01
Removed brmask CAM from ROB.

bkorpan:
Entries outside of head/tail range are implicitly invalid.

### DIFF
--- a/src/main/scala/exu/rob.scala
+++ b/src/main/scala/exu/rob.scala
@@ -245,7 +245,8 @@ class Rob(
    val full = Wire(Bool())
    val empty = Wire(Bool())
 
-   val r_tail_bump = Reg(Bool())
+   // Did the tail pointer move (from a row dispatch, mispredict, or completed rollback) on the last cycle?
+   val r_tail_moved = Reg(Bool())
 
    val will_commit         = Wire(Vec(width, Bool()))
    val can_commit          = Wire(Vec(width, Bool()))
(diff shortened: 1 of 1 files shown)


jerryz123:
Why is `r_tail_bump` needed now?

src/main/scala/exu/rob.scala
```diff
@@ -287,22 +289,37 @@ class Rob(
...
+         rob_fflags(disp_idx)    := 0.U
+         rob_uop(disp_idx).stat_brjmp_mispredicted := false.B
 
-         assert (rob_val(rob_tail) === false.B, "[rob] overwriting a valid entry.")
+         // Entries outside of head/tail range are now implicitly invalid.
+         //assert (rob_val(rob_tail) === false.B, "[rob] overwriting a valid entry.")
          assert ((io.enq_uops(w).rob_idx >> log2Ceil(width)) === rob_tail)
       }
-      .elsewhen (io.enq_valids.reduce(_|_) && !rob_val(rob_tail))
+      .elsewhen (io.brinfo.mispredict)
```
ccelio, line 342: why is mispredict lower priority than enq_valid? If these are mutually exclusive, then you should add an assert to verify that you are never enqueuing entries while a branch mispredict occurs.

bkorpan:
> 
> 
> Why is `r_tail_bump` needed now?

Need to make sure any dispatch bubbles are marked as invalid (any entry in the tail row which doesn't receive a valid entry on the first cycle after the tail increments), and avoid unrolling invalid entries which are falsely valid, if the rob_state transitions to s_rollback on the same cycle that the tail increments.

src/main/scala/exu/rob.scala
bkorpan: They should be mutually exclusive. You're right, an assert would make sense.

jerryz123:
> > Why is `r_tail_bump` needed now?
> 
> Need to make sure any dispatch bubbles are marked as invalid (any entry in the tail row which doesn't receive a valid entry on the first cycle after the tail increments), and avoid unrolling invalid entries which are falsely valid, if the rob_state transitions to s_rollback on the same cycle that the tail increments.

Can you add a comment for this when r_tail_bump is defined? This behavior is not obvious.

PR #214, opened 2019-03-01
Made fast wakeup broadcast to busytable optional.

bkorpan:


src/main/scala/exu/core.scala
```diff
@@ -669,13 +676,22 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
 
+   wu_idx = 0
    for ((renport, intport) <- rename_stage.io.int_wakeups zip int_wakeups)
    {
-      // Stop wakeup for bypassable children of spec-loads trying to issue during a ldMiss.
-      renport.valid :=
-         intport.valid &&
-         !(sxt_ldMiss && (intport.bits.uop.iw_p1_poisoned || intport.bits.uop.iw_p2_poisoned))
-      renport.bits := intport.bits
+      if (wu_idx < num_irf_write_ports + 1)
```
jerryz123, line 682: This looks more hacky than I'd like. 

Can we tag the ExeUnitResp bundles with a "is_fast_wakeup" field, and then dynamically do this assignment here? We can add some asserts to make sure we have the number of fast and slow wakeups we expect.

pentin-as:
enablePipelinedAllocation = true,
enableFastWakeupsToRename = false,

/home/pentin-as/riscv/00001_titan/ext/ip/00003_riscv/sim/gen/mm.riscv started
C0: reg block 4x5x6, cache block 24x25x24
C0: 24810 instructions
C0: 17906 cycles
C0: 28800 flops
C0: 1608 Mflops @ 1 GHz
C0: c[0][4] %f != %f
benchmark failure
 ... failed, error code = 3

I'll try to figure out why.

pentin-as:
enablePipelinedAllocation = false,
enableFastWakeupsToRename = false,
works
enablePipelinedAllocation = true,
enableFastWakeupsToRename = true,
does not work

This PR works. The problem in https://github.com/riscv-boom/riscv-boom/pull/215.

jerryz123 (changes requested):

PR #215, opened 2019-03-01
Added option for pipelined physical register allocation.

bkorpan:


pentin-as:
The virtual register t6 refers to the physical register 0x32.
screenshot from 2019-03-04 18-31-52
At this time, the virtual register a7 also starts referring to the physical register 0x32 and distorts the t6 data.
screenshot from 2019-03-04 18-33-56
From above a7, from below t6:
screenshot from 2019-03-04 18-35-47
If enablePipelinedAllocation = false, there is no problem.

bkorpan:
Interesting, thanks for finding this. What config were you using when this happened?

PR #218, opened 2019-03-14
[bpu] Misc Cleanup of the BPU

abejgonzalez:
There is no functional difference. 

Cleans up what parameters are used to instantiate a BPU.
Cleans up unused parameters.
Renames `BrPredictor` to `BoomBrPredictor` for a consistent name with `BoomBTB`
Enforces that there is 1 BPU instantiated when the enabled flag is set
Default to using a BPU (instead of a null BPU)
Other small changes (comments, code cleanup)


abejgonzalez:
Note: Since this ensures that there is only 1 BPU, this makes writing a mixin a bit "harder" in that you have to clear all other BPU's by doing a `None` to the prior one (although I think this is clearer to read and avoids someone forgetting the order of the `if` statement in `br-predictor.scala`)

jerryz123 (approved):

abejgonzalez:
I will merge this then!

PR #219, opened 2019-03-15
named fb/ftq rams | named issue queues | renamed regfile

abejgonzalez:
- Renamed RegFile from `Behavioral` to `Synthesizable` since "behavioral" normally indicates that the instance cannot be synthesized. 
- Named all issue units so that it is easier to read the output verilog signals
- Named FTQ and FB ram so that it is easier to read the output verilog signals

### DIFF
--- a/src/main/scala/exu/issue-units/issue-units.scala
+++ b/src/main/scala/exu/issue-units/issue-units.scala
@@ -52,12 +52,19 @@ class IssueUnits(num_wakeup_ports: Int)(implicit val p: Parameters)
 
    for (issueParam <- issueParams.filter(_.iqType != IQT_FP.litValue))
    {
-      val issueUnit = Module(new IssueUnitCollasping(issueParam, num_wakeup_ports))
+      val issueUnit = Module(new IssueUnitCollapsing(issueParam, num_wakeup_ports))
 
       // name the issue units
       if (issueParam.iqType == IQT_INT.litValue)
       {
-        issueUnit.suggestName("int_issue_unit")
+        if (usingUnifiedMemIntIQs)
+        {
+          issueUnit.suggestName("intmem_issue_unit")
+        }
+        else
+        {
+          issueUnit.suggestName("int_issue_unit")
+        }
       }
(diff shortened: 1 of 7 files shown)


src/main/scala/exu/issue-units/issue-units.scala
```diff
@@ -50,9 +50,21 @@ class IssueUnits(num_wakeup_ports: Int)(implicit val p: Parameters)
 
    require (enableAgePriorityIssue) // unordered is currently unsupported.
 
-   for (issParam <- issueParams.filter(_.iqType != IQT_FP.litValue))
+   for (issueParam <- issueParams.filter(_.iqType != IQT_FP.litValue))
    {
-      iss_units += Module(new IssueUnitCollasping(issParam, num_wakeup_ports))
+      val issueUnit = Module(new IssueUnitCollasping(issueParam, num_wakeup_ports))
```
jerryz123: Since you are PRing name changes anyways, can you correct Collasping to Collapsing

src/main/scala/exu/issue-units/issue-units.scala
```diff
@@ -50,9 +50,21 @@ class IssueUnits(num_wakeup_ports: Int)(implicit val p: Parameters)
...
+
+      // name the issue units
+      if (issueParam.iqType == IQT_INT.litValue)
+      {
+        issueUnit.suggestName("int_issue_unit")
+      }
+      else if (issueParam.iqType == IQT_MEM.litValue)
+      {
+        issueUnit.suggestName("mem_issue_unit")
+      }
```
jerryz123, line 77: We also have the MEMINT IQ for unified, might as well handle that here too.

jerryz123 (changes requested):

src/main/scala/exu/issue-units/issue-units.scala
abejgonzalez: Wow. I didn't even notice that. Good catch!

abejgonzalez:
Fixed email/username that GitHub pointed to in the rebase

jerryz123 (approved):

PR #220, opened 2019-03-15
Revert "Do not connect the PTW if not usingPTW"

abejgonzalez:
Reverts riscv-boom/riscv-boom#198

The issue is that when synthesizing without PTW's, it optimizes out all of the core. So I am going to revert this until we know how to properly disconnect the PTW's.


jerryz123 (approved):

aswaterman:
The PTW is used to distribute `mstatus` and the PMPs around the core, which are needed even if virtual memory is disabled.  (It's analogous to the TLBs being needed even if virtual memory is disabled, since they contain the PMP/PMA checkers.)

PR #221, opened 2019-03-15
Preliminary ROCC support

jerryz123:
Currently this supports a subset of the RoCC interface. Should be enough for Hwacha.


src/main/scala/exu/core.scala
```diff
@@ -1477,44 +1504,23 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
-   io.rocc.mem.s2_nack := false.B
-   io.rocc.mem.s2_nack_cause_raw := false.B
-   io.rocc.mem.s2_xcpt.ma.st := false.B
-   io.rocc.mem.s2_xcpt.ma.ld := false.B
-
-   // Wire off other unused CoreIO signals.
-   io.fpu <> DontCare
-   io.fpu.valid := false.B
-   io.fpu.inst := 0.U
+   if (usingRoCC)
```
ccelio, line 1535: What happens if !usingRoCC? Do you still tie off the io.rocc signals?

jerryz123: No, but is it necessary to? They don't drive anything in that case.

jerryz123:
This seems to work with Hwacha.

src/main/scala/exu/core.scala
ccelio: I don't think it's good practice to leave I/Os dangling---and I've occasionally be surprised when I thought IOs would get trimmed and they didn't (fields within the MicroOp has bitten me here).

Related, I _strongly_ recommend that all modules, including BoomCore, add `override val compileOptions = chisel3.core.ExplicitCompileOptions.NotStrict.copy(explicitInvalidate = true)` to catch undriven I/Os. That will force you to specify DontCare if you want to leave stuff undriven. Importing `chisel3` isn't strong enough here.

jerryz123:
@bkorpan I reverted the change to disambiguate rob ordering by head, instead of tail. Instead I fixed the bug where `rob_tail_idx` is inaccurate. Now `rob_tail_idx` has entry granularity, rather than row granularity.

src/main/scala/exu/rob.scala
```diff
@@ -799,26 +818,36 @@ class Rob(
...
    {
-      rob_tail := WrapInc(rob_tail, num_rob_rows)
-      rob_enq := true.B
+      rob_tail     := WrapInc(rob_tail, NUM_ROB_ROWS)
+      rob_tail_lsb := 0.U
+      rob_enq      := true.B
+   }
+   .elsewhen (io.enq_valids.asUInt =/= 0.U && io.enq_partial_stall)
+   {
+      rob_tail_lsb := PriorityEncoder(~io.enq_valids.asUInt)
```
jerryz123: @ccelio @bkorpan Any thoughts on how the tail bank idx bits are calculated here?

docs/sections/reorder-buffer.rst
```diff
@@ -193,3 +193,17 @@ Load/Store Unit, and are treated as exceptions in the BOOM pipeline
 .. [3]
    The tradeoff here is between longer latencies on exceptions versus an
    increase in area and wiring.
+
+Point of No Return (PNR)
+------------------------
+
+The point-of-no-return head runs ahead of the ROB commit head, marking the
```
abejgonzalez: "marking the next instruction which might not commit and cause misspeculation"

src/main/scala/common/rocc.scala
```diff
@@ -0,0 +1,69 @@
+package boom.common
```
abejgonzalez, line 15: Can we add the proper heading here.

docs/sections/execution-stages.rst
```diff
@@ -322,3 +322,35 @@ an exception as part of executing a CSR instruction (e.g., a syscall).
 
 .. [7]
    The CSRFile is a Rocket component.
+
```
abejgonzalez, line 325: Is it possible to get a block diagram of the connection to the ROB, queues, etc? (Not going to block the PR on this though)

docs/sections/reorder-buffer.rst
abejgonzalez, line 200: The sentence sounds weird? "not commit and cause misspeculation"

docs/sections/execution-stages.rst
jerryz123: I added the RoCC thing to the basic block diagram we have on lucidchart

src/main/scala/common/rocc.scala
```diff
@@ -0,0 +1,69 @@
...
+import chisel3._
+import chisel3.util.{RRArbiter, Queue}
+
+import freechips.rocketchip.rocket._
+import freechips.rocketchip.tile._
+import freechips.rocketchip.util._
+
+import boom.lsu._
+
+/*
```
abejgonzalez: Are we adhering to the following syntax for multi-line comments

```
/**
 * This is a comment
 */
```

src/main/scala/exu/fp-pipeline.scala
```diff
@@ -168,7 +168,6 @@ class FpPipeline(implicit p: Parameters) extends BoomModule()(p) with tile.HasFP
    //-------------------------------------------------------------
 
    exe_units.map(_.io.brinfo := io.brinfo)
-   exe_units.map(_.io.com_exception := io.flush_pipeline)
```
abejgonzalez, line 171: Why remove this?

src/main/scala/exu/rob.scala
```diff
@@ -230,25 +232,27 @@ class Rob(
...
+   val rob_tail     = RegInit(0.U(log2Ceil(NUM_ROB_ROWS).W))
+   val rob_tail_lsb = RegInit(0.U((1 max log2Ceil(width)).W))
+   val rob_tail_idx = if (width == 1) rob_tail else Cat(rob_tail, rob_tail_lsb)
 
-   val maybe_full = RegInit(false.B)
-   val full = Wire(Bool())
-   val empty = Wire(Bool())
+   val rob_pnr      = RegInit(0.U(log2Ceil(NUM_ROB_ROWS).W))
+   val rob_pnr_lsb  = RegInit(0.U((1 max log2Ceil(width)).W))
+   val rob_pnr_idx  = if (width == 1) rob_pnr  else Cat(rob_pnr , rob_pnr_lsb)
```
abejgonzalez, line 245: Nitpick: Spacing doesn't match rest of coding style

src/main/scala/exu/fp-pipeline.scala
jerryz123: I modified the execution unit IO such that only execution units which require the com_exception signal will have it in the IO

In general I'm trying to move towards IO bundles without extraneous signals.

src/main/scala/exu/rob.scala
jerryz123: The spacing is to match `rob_head_idx` and `rob_tail_idx`.

src/main/scala/util/util.scala
```diff
@@ -353,6 +342,19 @@ object AgePriorityEncoder
   }
 }
 
+// is i0 older than i1? (closest to zero). Provide the tail_ptr to the
```
abejgonzalez: Put the comment inside the `/** */` comment

src/main/scala/exu/execution-units/rocc.scala
```diff
@@ -0,0 +1,209 @@
+//******************************************************************************
+// Copyright (c) 2013 - 2018, The Regents of the University of California (Regents).
```
abejgonzalez: 2019

src/main/scala/exu/execution-units/rocc.scala
```diff
@@ -0,0 +1,209 @@
...
+
+package boom.exu
+
+import chisel3._
+import chisel3.util._
+import chisel3.experimental.dontTouch
+
+import freechips.rocketchip.config.Parameters
+import freechips.rocketchip.tile.{RoCCCoreIO, RoCCInstruction}
+import freechips.rocketchip.rocket._
```
abejgonzalez, line 23: Space between rocket and boom imports

src/main/scala/exu/execution-units/rocc.scala
```diff
@@ -0,0 +1,209 @@
...
+  val rxq_idx          = Output(UInt(log2Ceil(NUM_RXQ_ENTRIES).W))
+  val rob_pnr_idx      = Input(UInt(ROB_ADDR_SZ.W))
+  val rob_tail_idx     = Input(UInt(ROB_ADDR_SZ.W))
+
+  val rocc             = Flipped(new RoCCCoreIO)
+}
+
+/**
+ * IO bundle representing the different signals to interact with the RoCC
+ *
```
abejgonzalez: Extra space

docs/sections/execution-stages.rst
abejgonzalez: Update in the docs then?

src/main/scala/exu/fp-pipeline.scala
abejgonzalez: I see. I think this makes sense, although, Im not fully on board with using `null` to create specific bundles. For now, I think this is fine. The alternative is to use the `Some/None/etc` construct that Scala provides.

src/main/scala/exu/rob.scala
abejgonzalez: But those vals are a couple lines away. I think the better style is to match the things below and above. So in this case, there is no similar val above or below, so you revert to the normal spacing.

abejgonzalez (approved):

PR #222, opened 2019-03-16
[rob] Get the PC from the FTQ when enqueuing exceptions to the ROB

### LINKED ISSUE #62: Failing ma_fetch test
The `template` branch (used by https://github.com/ccelio/boom-template/commit/98312e162be9673116ed2b0db9191f149006528a) now fails the `ma_fetch` test in riscv-tests (the newer tests explore more cases).

jerryz123:
Resolves #62

The ROB should receive the PC of enqueuing uops when one of those uops causes an exception. 
When the frontend sends an excepting uop, wait for the ROB to empty first. Then get the PC from the FTQ.

### DIFF
--- a/src/main/scala/exu/core.scala
+++ b/src/main/scala/exu/core.scala
@@ -888,8 +888,23 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
          Causes.store_page_fault.U,
          Causes.fetch_page_fault.U)
 
-   csr.io.tval := Mux(tval_valid,
-      encodeVirtualAddress(rob.io.com_xcpt.bits.badvaddr, rob.io.com_xcpt.bits.badvaddr), 0.U)
+
+   // If the fetch access or page fault is on a 32b instruction misaligned across a page, (ex: 0x4ffe),
+   // the first exception needs to see badaddr = 0x4ffe, the second exception needs to see 0x5000.
+   // This is somewhat confusing, because typically we subtract 2 for edge instructions, here we add 2
+   // to get back the aligned address (which we need here for the second exception)
+   val tval_addr_pc = Mux((rob.io.com_xcpt.bits.uopc === uopJAL &&
+                           rob.io.com_xcpt.bits.cause === Causes.misaligned_fetch.U),
+                          ComputeJALTarget(csr.io.pc, ExpandRVC(rob.io.com_xcpt.bits.inst), xLen),
+                          csr.io.pc + Mux(rob.io.com_xcpt.bits.edge_inst, 2.U, 0.U))
+   val tval_addr = Mux(csr.io.cause.isOneOf(Causes.misaligned_fetch.U,
+                                            Causes.fetch_access.U,
+                                            Causes.fetch_page_fault.U),
+                       tval_addr_pc,
+                       rob.io.com_xcpt.bits.badvaddr)
+
+
+   csr.io.tval := Mux(tval_valid, encodeVirtualAddress(tval_addr, tval_addr), 0.U)
 
    // TODO move this function to some central location (since this is used elsewhere).
    def encodeVirtualAddress(a0: UInt, ea: UInt) = if (vaddrBitsExtended == vaddrBits)
@@ -1071,9 +1086,9 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
    //-------------------------------------------------------------
 
    // Dispatch
-   rob.io.enq_valids := rename_stage.io.ren1_mask
-   rob.io.enq_uops   := rename_stage.io.ren1_uops
-   rob.io.enq_new_packet := dec_finished_mask === 0.U
+   rob.io.enq_valids        := rename_stage.io.ren1_mask
+   rob.io.enq_uops          := rename_stage.io.ren1_uops
+   rob.io.enq_new_packet    := dec_finished_mask === 0.U
    rob.io.enq_partial_stall := dec_last_inst_was_stalled // TODO come up with better ROB compacting scheme.
    rob.io.debug_tsc := debug_tsc_reg
    rob.io.csr_stall := csr.io.csr_stall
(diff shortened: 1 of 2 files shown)


src/main/scala/exu/core.scala
```diff
@@ -525,9 +527,18 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
                         || rob.io.flush.valid
                         || dec_stall_next_inst
                         || (dec_uops(w).is_fencei && !lsu.io.lsu_fencei_rdy)
+                        || (dec_uops(w).exception
```
ccelio: i don't think you can do this -- the exception signal will be very late arriving and the `stall_me` signal is driving a serialized chain of uops.

ccelio (changes requested):
At a high-level, I don't see why you need to read the exception PC during decode; you can trivially grab it during commit. In fact, you should already have that commit port on the FTQ's PCFile because the CSRFile needs the commit PC (esp. for future RISC-V Standard Trace port stuff) and so does updating the predictors.

src/main/scala/exu/core.scala
```diff
@@ -888,8 +888,23 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
          Causes.fetch_page_fault.U)
 
-   csr.io.tval := Mux(tval_valid,
-      encodeVirtualAddress(rob.io.com_xcpt.bits.badvaddr, rob.io.com_xcpt.bits.badvaddr), 0.U)
+
+   // If the fetch access or page fault is on a 32b instruction misaligned across a page, (ex: 0x4ffe),
+   // the first exception needs to see badaddr = 0x4ffe, the second exception needs to see 0x5000.
+   // This is somewhat confusing, because typically we subtract 2 for edge instructions, here we add 2
+   // to get back the aligned address (which we need here for the second exception)
+   val tval_addr_pc = Mux((rob.io.com_xcpt.bits.uopc === uopJAL &&
```
ccelio, line 896: In the ROB, you don't have access to uop.bits.uopc nor uop.bits.inst, so you can't compute the trap value PC in this manner.

src/main/scala/exu/core.scala
```diff
@@ -888,8 +888,23 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
+                           rob.io.com_xcpt.bits.cause === Causes.misaligned_fetch.U),
+                          ComputeJALTarget(csr.io.pc, ExpandRVC(rob.io.com_xcpt.bits.inst), xLen),
```
ccelio, line 898: The `inst` bits that enter the backend are already expanded; and they are thrown away before entering the ROB, so you can't compute the JAL target in this manner.

ccelio (changes requested):
The trap value comes from two possibly places: 1) the ROB's badvaddr/tval (breakpoints, misaligned loads/stores, etc.) or 2) the FTQ's fetch-address (which gets passed to csr.io.pc). 

The csr.io.pc requires some clean-up to deduce what the actual PC of the current head-of-the-ROB is; if the instruction is split across FTQ entries (because it required two cycles to fetch all of the bits), the fetch-address coming from the FTQ needs to be readjusted to point to the first-half of the instruction; otherwise you simply add the low-order bits stored in the ROB to the fetch-address coming from the FTQ to get the exact `pc` of your particular, excepting instruction. 

The final corner case that gets weird is an instruction page fault for "split" instructions; you don't want to point to the PC of the failing instruction, you want to point to the VM page.

None of this requires recomputing jump/branch targets using instruction bits or opcodes; you only need to know the FTQ's fetch-address, the low-order PC bits, the exception cause, and is_split for the excepting instruction.

jerryz123:
My

> The trap value comes from two possibly places: 1) the ROB's badvaddr/tval (breakpoints, misaligned loads/stores, etc.) or 2) the FTQ's fetch-address (which gets passed to csr.io.pc).
> 
> The csr.io.pc requires some clean-up to deduce what the actual PC of the current head-of-the-ROB is; if the instruction is split across FTQ entries (because it required two cycles to fetch all of the bits), the fetch-address coming from the FTQ needs to be readjusted to point to the first-half of the instruction; otherwise you simply add the low-order bits stored in the ROB to the fetch-address coming from the FTQ to get the exact `pc` of your particular, excepting instruction.
> 
> The final corner case that gets weird is an instruction page fault for "split" instructions; you don't want to point to the PC of the failing instruction, you want to point to the VM page.
> 
> None of this requires recomputing jump/branch targets using instruction bits or opcodes; you only need to know the FTQ's fetch-address, the low-order PC bits, the exception cause, and is_split for the excepting instruction.

In the case of !usingCompressed, wouldn't JAL instructions potentially throw misaligned_fetch exceptions? That is the case which is annoying, the target address needs to be regenerated using the inst bits and pc.

The case of a 32b instruction split across pages in RVC is already handled in the way edge_inst is set.

aswaterman:
The rocket core sets tval to 0 on misaligned-fetch exceptions for similar reasons.

The privileged spec says, "The hardware platform will specify which exceptions must set mtval informatively and which may unconditionally set it to zero."  In the absence of a platform spec, we've taken the interpretation that, since misaligned-fetch exceptions are not resumable, providing tval isn't especially useful to software, and so it's not worth going out of our way to support it.

ccelio:
Yes, my recommendation is to follow what Rocket does and set `tval=0` on misaligned-fetch. Technically you could shove the bad JAL's target into the next FTQ entry (it may already do this) but it's not worth the headache of accelerating an unrecoverable error.

jerryz123:
What was the issue last year, which prompted you to add the hack? I see github issues referencing expanded ma_fetch test cases. It seems like setting tval=0 works with the current riscv-tests?

ccelio:
I don't remember; but I think I was trying to set tval to something and was getting it wrong, and didn't realize I could escape out by setting it to tval=0.

PR #226, opened 2019-03-21
[WIP] Style Guide

abejgonzalez:
This PR is to add a file to the repo serving as the baseline of how we want to format the code. The goal of this is to make reading the codebase easier for other people and to be able to point to style guide for PR's. It will say open for any/all to help define what the style should be. Please leave comments / concerns, etc.

### DIFF
--- a/CONTRIB_AND_STYLE.md
+++ b/CONTRIB_AND_STYLE.md
@@ -28,11 +28,14 @@ We have the following coding guidelines:
     * E.g. `RegisterFile`
 * If the variable is in a particular stage, please put the stage cycle it is associated with
     * E.g. `val f2_valid = Wire(...)`
-* Braces for control structures (`for`, `while`, `if`, `when`, etc...) should be on the same line as the condition
+* Braces for control structures (`for`, `while`, `if`, `when`, etc...) and functions should be on the same line as the condition
     * E.g. `for (i <- something) {`
-* Braces for classes should be on the next line
+    * E.g. `def foo(): Type = {`
+* Braces for classes and objects should be on the next line
     * ```
       class SomeModule extends Module
+      {
+      object SomeObject
       {
       ```
 * Things such as `else`, `.otherwise`, etc should be on the same line as the closing bracket of the previous scope
--- a/CONTRIB_AND_STYLE.md
+++ b/CONTRIB_AND_STYLE.md
@@ -35,33 +35,33 @@ In addition we have the following coding guidelines:
 * Braces for control structures (`for`, `while`, `if`, `when`, etc...) should be on the same line as the condition
     * E.g. `for (i <- something){`
 * Braces for classes should be on the next line
-    * E.g. ```
-           class SomeModule extends Module
-           {
-           ```
+    * ```
+      class SomeModule extends Module
+      {
+      ```
 * Braces for things such as `else`, `.otherwise`, etc should be on the next line
-    * E.g. ```
-           if (...){
-             ...
-           }
-           else{
-           ```
+    * ```
+      if (...){
+        ...
+      }
+      else{
+      ```
 * Indentation should be two spaces...
-    * E.g. ```
-           if (...){
-             if (...){
-               if (...){
-                 ...
-               }
-             }
-           }
+    * ```
+      if (...){
+        if (...){
+          if (...){
+            ...
+          }
+        }
+      }
            ```
 * Multiline comments within the code should be indented on the second line associated with it
-    * E.g. ```
-           // this is a start of a comment that
-           //   overflows to the next line that
-           //   overflows for the final time
-           ```
+    * ```
+      // this is a start of a comment that
+      //   overflows to the next line that
+      //   overflows for the final time
+      ```
 * To denote a new section of the code that needs to be separated with a comment, use the follwing format
     * ```
       //----------------------------
--- a/CONTRIB_AND_STYLE.md
+++ b/CONTRIB_AND_STYLE.md
@@ -24,11 +24,15 @@ In addition we have the following coding guidelines:
     * E.g. `local_variable_name`
 * Classes should be in camel case with the first letter capitalized
     * E.g. `RegisterFile`
-* Registers should be labeled with a `r_` before the name
-    * E.g. `val r_f2_valid = Reg(...)`
+* Registers should be labeled with a `_r` after the name
+    * E.g. `val f2_valid_r = Reg(...)`
+* Queues should be labeled with a `_q` after the name
+    * E.g. `val imem_resp_q = Queue(...)`
 * If the variable is in a particular stage, please put the stage cycle it is associated with
     * E.g. `val f2_valid = Wire(...)`
-* Braces for control structures (`for`, `while`, `if`, etc...) should be on the same line as the condition
+* Braces for control structures (`for`, `while`, `if`, `when`, etc...) should be on the same line as the condition
     * E.g. `for (i <- something){`
 * Braces for classes should be on the next line
     * E.g. `class SomeModule extends Module\n{`
+* Braces for things such as `else`, `.otherwise`, etc should be on the next line
+    * E.g. `if (...){\n...\n}\nelse{`
--- a/CONTRIB_AND_STYLE.md
+++ b/CONTRIB_AND_STYLE.md
@@ -15,19 +15,17 @@ We also recommend that you try to cut up contributions into smaller, digestable 
 
 **Style Guide**
 
-You can invoke `make checkstyle` to verify that your changes respect the style guide.
-In addition we have the following coding guidelines:
+You can invoke `make checkstyle` to verify that some of your changes respect the style guide.
+However, checkstyle does not cover all cases of the code.
 
-* Scala global variables should be in capital snake case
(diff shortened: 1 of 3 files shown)


abejgonzalez:
One thing to note about camel case...

It might be hard to read verilog without the `w/r/q` appended to the signal. This could lead to names looking very similar. However, this only comes up if one name is a subset of another and the smaller one is something like a bundle/vec, etc.

CONTRIB_AND_STYLE.md
```diff
@@ -0,0 +1,70 @@
...
+* Global variables should be in capital snake case
+    * E.g. `GLOBAL_VARIABLE_NAME`
+* Local variables should be in lowercase snake case
+    * E.g. `local_variable_name`
+* Classes should be in camel case with the first letter capitalized
+    * E.g. `RegisterFile`
+* Registers should be labeled with a `_r` after the name
+    * E.g. `val f2_valid_r = Reg(...)`
+* Wires should be labeled with a `_w` after the name
+    * E.g. `val f2_valid_w = Wire(...)`
```
jerryz123: Variables without a `_r` are implicitly wires. I don't see a need to explicitly name these

CONTRIB_AND_STYLE.md
```diff
@@ -0,0 +1,70 @@
...
+* Queues should be labeled with a `_q` after the name
+    * E.g. `val imem_resp_q = Queue(...)`
```
jerryz123: This isn't really common enough of an idiom to require its own standard naming.

CONTRIB_AND_STYLE.md
```diff
@@ -0,0 +1,70 @@
...
+        ...
+      }
+      else{
+      ```
+* Indentation should be two spaces...
+    * ```
+      if (...){
+        if (...){
+          if (...){
+            ...
```
jerryz123, line 54: I don't have a strong opinion on brace-on-same-line vs brace-on-next-line. I think we should go with whatever most of the code is currently doing, which appears to be brace-on-next-line.

CONTRIB_AND_STYLE.md
```diff
@@ -0,0 +1,70 @@
...
+
+**Style Guide**
+
+You can invoke `make checkstyle` to verify that your changes respect the style guide.
+In addition we have the following coding guidelines:
+
+* Global variables should be in capital snake case
+    * E.g. `GLOBAL_VARIABLE_NAME`
+* Local variables should be in lowercase snake case
+    * E.g. `local_variable_name`
```
jwright6323: I have no dog in this fight, but `snake_case` is notably against the chisel (and scala) style guide, which includes justification:
https://github.com/freechipsproject/chisel3/blob/master/doc/style.md

CONTRIB_AND_STYLE.md
abejgonzalez: Thats because I changed to be that way... that method takes up alot of extra space so thats why i prefer the same line bracket

CONTRIB_AND_STYLE.md
abejgonzalez: My comment above briefly mentions this... however, from speaking with the Berkeley BOOM crew, it sounds like we want to mainly keep with the snake case for readability in the Chisel. In the verilog it will be a bit harder to decipher... but as long as variables are subsets of each other... aka (`f2_valid` and `f2_valid_something`) I think this is fine.

CONTRIB_AND_STYLE.md
bkorpan: Don't we want a space between the condition brackets and scope brace?

CONTRIB_AND_STYLE.md
abejgonzalez: Im fine with getting rid of the `q` and `w` for now.

CONTRIB_AND_STYLE.md
abejgonzalez: For further clarification... I changed the `{` to be on a different line everywhere to get things consistent. However, now that we are looking closer at a particular style, we can look to change this.

@bkorpan The Google style guide seems to do this. I personally don't care about this. We can make it `if (...) {` if you'd like.

abejgonzalez:
Another thing that should be discussed is the difference between Chisel variables and Scala parameter variables. We can do both the same way but this makes distinguishing what is Scala and what is Chisel, a bit harder in the code. Maybe we have Chisel vars in snake_case and Scala params in camelCase?

CONTRIB_AND_STYLE.md
bkorpan: > 
> 
> For further clarification... I changed the `{` to be on a different line everywhere to get things consistent. However, now that we are looking closer at a particular style, we can look to change this.
> 
> @bkorpan The Google style guide seems to do this. I personally don't care about this. We can make it `if (...) {` if you'd like.

Yeah, I'd greatly prefer that.

bkorpan:
> 
> 
> Another thing that should be discussed is the difference between Chisel variables and Scala parameter variables. We can do both the same way but this makes distinguishing what is Scala and what is Chisel, a bit harder in the code. Maybe we have Chisel vars in snake_case and Scala params in camelCase?

Yeah, the thing that bugs me most right now is the naming of parameters/non-Chisel variables. I think the sensible options are:
- snake_case for local variables (same as Chisel nodes) and ALL_CAPS_SNAKE for implicitly passed in parameters (how it used to be I believe)
- camelCase for all Scala variables (this is partially the case now, the conversion process was started at some point)
- camelCase for local Scala variables and ALL_CAPS_SNAKE for implicit parameters... I'm not really a fan of mixing these, but it might be good to have the obvious distinction. This is more-or-less how it is right now.

abejgonzalez:
> > Another thing that should be discussed is the difference between Chisel variables and Scala parameter variables. We can do both the same way but this makes distinguishing what is Scala and what is Chisel, a bit harder in the code. Maybe we have Chisel vars in snake_case and Scala params in camelCase?
> 
> Yeah, the thing that bugs me most right now is the naming of parameters/non-Chisel variables. I think the sensible options are:
> 
>     * snake_case for local variables (same as Chisel nodes) and ALL_CAPS_SNAKE for implicitly passed in parameters (how it used to be I believe)
> 
>     * camelCase for all Scala variables (this is partially the case now, the conversion process was started at some point)
> 
>     * camelCase for local Scala variables and ALL_CAPS_SNAKE for implicit parameters... I'm not really a fan of mixing these, but it might be good to have the obvious distinction. This is more-or-less how it is right now.

I personally prefer the 3rd case because there is a clear distinction.

CONTRIB_AND_STYLE.md
```diff
@@ -0,0 +1,70 @@
...
+
+* Scala global variables should be in capital snake case
+    * E.g. `GLOBAL_SCALA_VARIABLE_NAME`
+* Scala local variables should be in lowercase camel case
+    * E.g. `localScalaVariableName`
+* Chisel variables should be in lowercase snake case
+    * E.g. `local_chisel_variable_name`
+* Classes should be in camel case with the first letter capitalized
+    * E.g. `RegisterFile`
+* Registers should be labeled with a `_r` after the name
```
abejgonzalez: Offline: Don't use _r instead try to rename all _r variables currently there into better variable names

CONTRIB_AND_STYLE.md
```diff
@@ -0,0 +1,70 @@
...
+* Braces for classes should be on the next line
+    * ```
+      class SomeModule extends Module
+      {
+      ```
+* Things such as `else`, `.otherwise`, etc should be on the next line
+    * ```
+      if (...) {
+        ...
+      }
```
abejgonzalez: Offline: Have the else be on the same line as the closing brace.

CONTRIB_AND_STYLE.md
```diff
@@ -0,0 +1,70 @@
...
+This will give you early feedback on the likelihood that we will be able to accept your changes, that you will not conflict with existing efforts, and will give you guidance that will help save you time on your own implementation.
+
+We also recommend that you try to cut up contributions into smaller, digestable pieces that are easier-to-review and easier-to-maintain.
+
+**Style Guide**
+
+You can invoke `make checkstyle` to verify that your changes respect the style guide.
+In addition we have the following coding guidelines:
+
+* Scala global variables should be in capital snake case
```
abejgonzalez: Offline: Have all scala vars be lowercase camelcase

jerryz123 (approved):

CONTRIB_AND_STYLE.md
```diff
@@ -0,0 +1,70 @@
...
+    * E.g. `localScalaVariableName`
+* Chisel variables should be in lowercase snake case
+    * E.g. `local_chisel_variable_name`
+* Classes should be in camel case with the first letter capitalized
+    * E.g. `RegisterFile`
+* If the variable is in a particular stage, please put the stage cycle it is associated with
+    * E.g. `val f2_valid = Wire(...)`
+* Braces for control structures (`for`, `while`, `if`, `when`, etc...) should be on the same line as the condition
+    * E.g. `for (i <- something) {`
+* Braces for classes should be on the next line
```
abejgonzalez: objects also, funcs should have { on same line

PR #227, opened 2019-03-21
[exu] Custom RegFile additions | RegFile UnitTests

abejgonzalez:
This is parameterizing the BROOM custom register file code that was written. Additionally, this is adding unit tests for the register files (this also serves as an example for other unit tests to be written). Though simple tests, this is the first step to getting more done. Tests can be run with `sbt test` to run all or with `sbt testOnly boom.tests.NameOfTestClass`.

Additionally, there is other small changes to comments / scalastyle checks.


src/test/scala/RegFileTest.scala
```diff
@@ -0,0 +1,300 @@
...
+    poke(c.io.write_ports(secondWPort).bits.data, 10.U) // some dummy value
+
+    try
+    {
+      // this should assert
+      step(1)
+    }
+    catch
+    {
+      case _: Throwable => println("This was supposed to throw an assertion")
```
jerryz123, line 196: LGTM, although I do think its odd that this is testing if an unsupported behavior is truly unsupported.

jerryz123 (approved):

src/test/scala/RegFileTest.scala
abejgonzalez: Oh. I forgot to remove this test... This was just to verify things on my end. I don't think this hurts to be here but I can remove it if need be.

Issue #170, opened 2019-01-27
Access exceptions caused by PTW are probably broken right now.


jerryz123:
The D$ TLB handles both PTW and core requests, and does not differentiate between either. To re-enable PMPs, #123 ignores access exceptions thrown by the D$ TLB, since core requests check the PMPs in the LSU TLB instead.

However, PTW requests may still fail a PMP check, and that exception needs to propagate.

Essentially, the D$ needs to differentiate between core and PTW requests somehow.

pentin-as:
Until you have done more closely L1D and LSU, I would like to try to make the following temporary correction:
 - TLB inside LSU only translate address;
 - D$ contains PMPChecker (instead LSU), which checks all requests (from LSU and PTW).

Additionally, I can say that PMP inside LSU contains one of the critical path that would be good to remove:
https://github.com/riscv-boom/riscv-boom/issues/224

What do you think about that?

jerryz123:
The issue is this would require forking the TLB and PMPChecker from Rocket, which currently combines them in one module. I'm not sure how much work this would be, and in any case forking would leave behind a lot of duplicate code between BOOM and Rocket. 

It also looks like it will be challenging to split up the functionality of PMPChecker from the TLB, since on refills, the PMPChecker marks bits in the TLB entries.

jerryz123:
Resolved in #275

PR #228, opened 2019-03-21
TLB completely removed from DCache

pentin-as:


jerryz123:
This looks good, the one issue is that PTW requests still need to check a TLB in case of PMP violations. #170.

We plan on redesigning the L1D$ more closely into the LSU in the future. Ideally there is 1 TLB/PMPChecker that all D$ requests go through.

PR #229, opened 2019-03-21
Rob changes

bkorpan:
Moved IsOlder comparison for wrapping queues to util.scala
The IsOlder comparison for tracking the oldest exception now uses rob_head to disambiguate order instead of rob_tail. Otherwise, an exception in the middle of a partially dispatched rob row (rob width > 2) could erroneously replace an older exception.

General refactor of PNR logic.
- Track "unsafeness" of entries rather than "safeness".
- Don't consider unconditional jumps to known addresses unsafe.
- No reason to check whether an ordering failure occurred before marking a load/store as safe
- Compute pnr_idx on instruction granularity to prevent potential deadlocks with atomic instructions dispatching to an upper column of a rob row, while a resource used by an instruction in a lower column of the same row is being held up by the PNR

Added fast PNR which can jump over large ranges of safe instructions in a cycle.

Let the ROB fill all of its entries.

### DIFF
--- a/src/main/scala/common/micro-op.scala
+++ b/src/main/scala/common/micro-op.scala
@@ -141,8 +141,8 @@ class MicroOp(implicit p: Parameters) extends BoomBundle()(p)
    val debug_events     = new DebugStageEvents
 
 
-   // Is it possible for this uop to not commit once it is in the ROB?
-   def may_xcpt         = (is_load || is_store || is_br_or_jmp || exception) && !(is_fence || is_fencei)
+   // Is it possible for this uop to misspeculate, preventing the commit of subsequent uops?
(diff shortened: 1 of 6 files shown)


jerryz123:
> * Compute pnr_idx on instruction granularity to prevent potential deadlocks with atomic instructions dispatching to an upper column of a rob row, while a resource used by an instruction in a lower column of the same row is being held up by the PNR

This actually bit me for a RoCC implementation. The PNR head idx will now advance past all valid entries, even though the end of a partially enqueued row?

jerryz123 (approved):

jerryz123:
@bkorpan I've approved this because I need to rebase RoCC (the only thing so far which uses PNR) onto this. 
Feel free to merge if you have no other changes in mind.

PR #230, opened 2019-03-22
Refactored LDQ/STQ entry allocation logic.

bkorpan:
LDQ and STQ will try to fill all entries. Decode is only stalled when at least one load or store cannot be allocated an entry.


jerryz123:
Can you fix the Rocc shim unit to use the newer isOlder

jerryz123:
You need to change rob_tail_idx to rob_head_idx for rocc module

jerryz123 (approved):

PR #231, opened 2019-03-22
Set tval=0 for ma_fetch from JALs. Also don't check for this if usingCompressed

jerryz123:


abejgonzalez (approved):

abejgonzalez:
#222 For reference.

PR #232, opened 2019-04-01
Cleanup | Printf Cleanup

abejgonzalez:
This PR is alot of cleanup to the printf's so that they are a bit more readable (at the cost of being a bit longer). Additionally, it has small changes to make the code more readable in certain places (BTB, Cfi/BrType, FetchController RVC). Overall, there should be no functional changes to the core.

I figured I would PR this and see if people like these changes since these are changes that I am currently using. If so, I can add/cleanup this PR more to put into `master`.


src/main/scala/bpu/btb/btb-sa.scala
```diff
@@ -84,67 +87,76 @@ class BTBsa(implicit p: Parameters) extends BoomBTB
...
+
+     // if multiple ways hit, clear the entry last read
+     when (clear_valid)
+     {
+        valids := valids.bitSet(clear_idx, false.B)
+     }
+
+     if (DEBUG_PRINTF)
+     {
+       printf("BTB-SA:\n")
```
abejgonzalez, line 136: I will disable this if this get merged once again. It pollutes too much of the log currently.

abejgonzalez:
I will reopen when more is done / cleaned.

PR #239, opened 2019-04-12
Feb. Rocket Bump

abejgonzalez:
Add things needed to compile `boom` with the latest `rocket-chip` found in `boom-template`


src/main/scala/exu/fp-pipeline.scala
```diff
@@ -198,7 +198,7 @@ class FpPipeline(implicit p: Parameters) extends BoomModule()(p) with tile.HasFP
       // Wakeup signal is sent on cycle S0, write is now delayed until end of S1,
       // but Issue happens on S1 and RegRead doesn't happen until S2 so we're safe.
       // (for regreadlatency >0).
-      fregfile.io.write_ports(0) <> WritePort(RegNext(ll_wbarb.io.out), FPREG_SZ, fLen+1)
+      fregfile.io.write_ports(0) <> WritePort(ll_wbarb.io.out, FPREG_SZ, fLen+1, delay_by_one=true)
```
jerryz123, line 201: Can you make the parameter name "latency", and `require(latency == 1 || latency == 0)` inside `WritePort`. 

Then you can just make this 

`fregfile.io.write_ports(0) <> WritePort(ll_wbarb.io.out, FPREG_SZ, fLen+1, regreadLatency)`

jerryz123 (changes requested):

src/main/scala/exu/fp-pipeline.scala
abejgonzalez: Will there ever be a case where the regreadlatency is > 1?  If so this, breaks down. While I agree this looks cleaner, it may break future functionality.

jerryz123: https://github.com/riscv-boom/riscv-boom/blob/7fd7fe5b7eeffb835c649cdbf27e4b836de9335a/src/main/scala/common/parameters.scala#L156

Turns out `regreadLatency = 0` is unsupported. I'm in favor of removing all references to `regreadLatency = 0`, and removing that parameterization option.

abejgonzalez: Hmm. Im not sure this discussion should be done on a bump then. I think we should punt this change to another PR where we can discuss this a bit more.

jerryz123: Sure, I'll open an issue

jerryz123 (approved):

PR #248, opened 2019-04-18
Refactor dispatch into module

jerryz123:


src/main/scala/common/parameters.scala
```diff
@@ -171,6 +168,11 @@ trait HasBoomCoreParameters extends freechips.rocketchip.tile.HasCoreParameters
    require (issueParams.count(_.iqType == IQT_MEM.litValue) == 1 || usingUnifiedMemIntIQs)
    require (issueParams.count(_.iqType == IQT_INT.litValue) == 1)
 
+   // Currently, require issue dispatch widths all equal coreWidth
+   issueParams.map(x => require(x.dispatchWidth == coreWidth))
+   // In future, relax this constraint
```
abejgonzalez: I would put a `TODO` here and comment out this line to make it apparent that the prior `require` is the dominating one.

src/main/scala/exu/core.scala
```diff
@@ -1329,31 +1301,32 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
       printf(") fin(%x)\n", dec_finished_mask)
       for (w <- 0 until coreWidth)
       {
+         val ren_uop = dispatcher.io.ren_uops(w).bits
```
abejgonzalez, line 1304: Ideally, we want this to match the dispatch width right according to the IQ not the coreWidth... right?

src/main/scala/exu/dispatch.scala
```diff
@@ -0,0 +1,64 @@
...
+   val io = IO(new DispatchIO)
+}
+
+/**
+  * This Dispatcher assumes worst case, all dispatched uops go to 1 issue queue
+  * This is equivalent to BOOMv2 behavior
+  */
+class BasicDispatcher(implicit p: Parameters) extends Dispatcher
+{
+   issueParams.map(x => require(x.dispatchWidth == coreWidth))
```
abejgonzalez: This is a redundant `require`. I would just keep this one and delete the one in params

src/main/scala/exu/dispatch.scala
```diff
@@ -0,0 +1,64 @@
...
+
+import chisel3._
+import chisel3.util._
+
+import freechips.rocketchip.config.Parameters
+
+import boom.common._
+import boom.util._
+
+class DispatchIO(implicit p: Parameters) extends BoomBundle()(p)
```
abejgonzalez, line 25: I would change all the `maps` in this section to be more descriptive with the `map` variables used. This should make it clearer to people trying to understand the code. Something like this:

`val ren_readys = io.dis_uops.map(iss_q => VecInit(iss_q.map(port => port.ready)).asUInt).reduce(_&_)`

src/main/scala/exu/dispatch.scala
```diff
@@ -0,0 +1,64 @@
...
+      io.ren_uops(w).ready := ren_readys(w)
+   }
+
+   for {i <- 0 until issueParams.size
+        w <- 0 until coreWidth} {
+      val issueParam = issueParams(i)
+      val dis        = io.dis_uops(i)
+
+
+      dis(w).valid := io.ren_uops(w).valid && (io.ren_uops(w).bits.iqtype & issueParam.iqType.U) =/= 0.U
```
abejgonzalez: I would put a paren. around the IQ type match.

src/main/scala/exu/dispatch.scala
```diff
@@ -0,0 +1,64 @@
...
+   {
+      io.ren_uops(w).ready := ren_readys(w)
+   }
+
+   for {i <- 0 until issueParams.size
+        w <- 0 until coreWidth} {
+      val issueParam = issueParams(i)
+      val dis        = io.dis_uops(i)
+
+
```
abejgonzalez: Extra space

src/main/scala/exu/dispatch.scala
```diff
@@ -0,0 +1,64 @@
...
+      dis(w).valid := io.ren_uops(w).valid && (io.ren_uops(w).bits.iqtype & issueParam.iqType.U) =/= 0.U
+      dis(w).bits  := io.ren_uops(w).bits
+
```
abejgonzalez: Extra space

src/main/scala/exu/core.scala
jerryz123: No. Dispatcher accepts incoming uops from rename of width `coreWidth`, and redirects them to issue queue ios each of width `dispatchWidth`

src/main/scala/exu/dispatch.scala
jerryz123: The one in params checks that we only support the BasicDispatcher.

In the future, we'll be able to choose between dispatchers. The BasicDispatcher checks that it can only be used in the case dispatchWidth == coreWidth.

src/main/scala/exu/dispatch.scala
```diff
@@ -42,9 +42,9 @@ abstract class Dispatcher(implicit p: Parameters) extends BoomModule()(p)
   */
 class BasicDispatcher(implicit p: Parameters) extends Dispatcher
 {
-   issueParams.map(x => require(x.dispatchWidth == coreWidth))
+   issueParams.map(ip => require(ip.dispatchWidth == coreWidth))
 
-   val ren_readys = io.dis_uops.map(x => VecInit(x.map(u => u.ready)).asUInt).reduce(_&_)
+   val ren_readys = io.dis_uops.map(dis => VecInit(dis.map(u => u.ready)).asUInt).reduce(_&_)
```
abejgonzalez: If you are going to leave it this short `dis` `u` I would add a comment here explaining what is happening (I prefer to expand the variable names though)

jerryz123: These variable names are short for the same reason that we write `uops.map(_.valid)` or `uops.map(u=>u.valid)` instead of `uops.map(uop => uop.valid)`. If you understand what `io.dis_uops` is, then you understand what these map expressions are doing.

In general variables that are used inside `map` statements should be concise, since the purpose of that variable is self-evident, and this allows the map statement to fit in one line.

abejgonzalez (approved):

PR #255, opened 2019-04-25
Issue Refactor

bkorpan:
Refactoring of issue slot logic. Wakeups are performed after slot entries are moved, elmiminating additional wakeup logic in the rename stage.


abejgonzalez (approved):
Seems fine to me on a quick review.

jerryz123:
Can you verify this on Fedora? Not just for these changes, but for several previous ones which we didn't verify thoroughly.

PR #256, opened 2019-05-01
Printf changes

abejgonzalez:
Clean up existing `printf`s in the code so that you dont need as much `Mux` logic everywhere. Also, did a quick pass on replacing stray `toBool` with `asBools`


src/main/scala/exu/core.scala
```diff
@@ -1172,144 +1172,108 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
-             Mux(csr.io.status.prv === (0x1).U, Str("S"),  //2 is H
-                                                Str("?"))))
-             )
+      val robTypeStrs = RobTypeToChars(rob.io.debug.state)
+      printf("ROB:\n")
+      printf("    (State:%c%c%c Rdy:%c LAQFull:%c STQFull:%c Flush:%c BMskFull:%c DShimRdy:%c) BMsk:0x%x Mode:%c\n",
+         robTypeStrs(0),
+         robTypeStrs(1),
+         robTypeStrs(2),
+         BoolToChar(                rob.io.ready, '_', '!'),
```
jerryz123: Is there a reason for this spacing?

src/main/scala/util/util.scala
```diff
@@ -461,3 +462,68 @@ class BranchKillableQueue[T <: boom.common.HasBoomUOP](gen: T, entries: Int)
...
+   * @param trueChar Scala Char if bool is true
+   * @param falseChar Scala Char if bool is false
+   * @return UInt ASCII Char for "trueChar" or "falseChar"
+   */
+  def apply(c_bool: Bool, trueChar: Char, falseChar: Char = '-'): UInt = {
+    Mux(c_bool, Str(trueChar), Str(falseChar))
+  }
+}
+
+object CfiTypeToChars
```
jerryz123, line 486: Why not make `strings` an argument, so you can collapse these three objects into one?

jerryz123 (changes requested):

src/main/scala/exu/issue-units/issue-unit.scala
```diff
@@ -175,36 +176,35 @@ abstract class IssueUnit(
...
+        BoolToChar(       issue_slots(i).valid, 'V'),
+        BoolToChar(     issue_slots(i).request, 'R'),
+        BoolToChar(issue_slots(i).in_uop.valid, 'W'),
+        BoolToChar(    issue_slots(i).debug.p1, '!'),
+        BoolToChar(    issue_slots(i).debug.p2, '!'),
+        BoolToChar(    issue_slots(i).debug.p3, '!'),
+        Mux(issue_slots(i).uop.dst_rtype === RT_FIX, Str("X"),
+            Mux(issue_slots(i).uop.dst_rtype === RT_X, Str("-"),
+                Mux(issue_slots(i).uop.dst_rtype === RT_FLT, Str("f"),
+                    Mux(issue_slots(i).uop.dst_rtype === RT_PAS, Str("C"), Str("?"))))),
```
bkorpan: Do we want these cascaded muxes indented by 4's?

src/main/scala/exu/issue-units/issue-unit.scala
```diff
@@ -175,36 +176,35 @@ abstract class IssueUnit(
...
+        issue_slots(i).uop.pop2,
+        issue_slots(i).uop.pop3,
+        issue_slots(i).uop.pc(31,0),
+        issue_slots(i).uop.debug_inst,
+        issue_slots(i).uop.uopc,
+        issue_slots(i).uop.rob_idx,
+        issue_slots(i).uop.br_mask,
+        issue_slots(i).uop.imm_packed)
     }
-    printf("-----------------------------------------------------------------------------------------\n")
```
bkorpan, line 207: We don't want this divider anymore?

src/main/scala/exu/core.scala
```diff
@@ -1172,144 +1172,108 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
-
-    val fpr_to_string =
-            VecInit( Str("ft0"), Str("ft1"), Str("ft2"), Str("ft3"),
-                 Str("ft4"), Str("ft5"), Str("ft6"), Str("ft7"),
-                 Str("fs0"), Str("fs1"), Str("fa0"), Str("fa1"),
-                 Str("fa2"), Str("fa3"), Str("fa4"), Str("fa5"),
-                 Str("fa6"), Str("fa7"), Str("fs2"), Str("fs3"),
-                 Str("fs4"), Str("fs5"), Str("fs6"), Str("fs7"),
-                 Str("fs8"), Str("fs9"), Str("fs10"), Str("fs11"),
-                 Str("ft8"), Str("ft9"), Str("ft10"), Str("ft11"))
```
bkorpan, line 1312: This isn't used anywhere? Might be useful to turn this into a function and put it in util.scala

bkorpan (changes requested):

src/main/scala/exu/core.scala
abejgonzalez: The thought was that this would make it a bit easier to read... but I can revert it if people care.

src/main/scala/util/util.scala
abejgonzalez: I think that makes it look worse and spreads out information across the places this function is called. For ex. if there is a change to the `cfi_type` then you have to change `strings` everywhere the function is called...

Im assuming you mean call the function like this?: `TypeToChars(Seq(...), type)`

src/main/scala/exu/core.scala
abejgonzalez: I will fix the bad spacing here though. Let me know if you want to remove the spacing in general

src/main/scala/exu/core.scala
abejgonzalez: This isn't used anywhere... however I can create a function of it in case it is used later.

src/main/scala/exu/issue-units/issue-unit.scala
abejgonzalez: Solved offline: Cascade by 2

bkorpan (approved):

src/main/scala/exu/issue-units/issue-unit.scala
abejgonzalez: No other place in the code has this line. I think this line was initially put since the FP Iss Slot `printf` was the last thing printed out. However, if we add more `printf`'s in more spots, this line gradually becomes interspersed with other `printfs` confusing things a bit. Thats why I removed it, instead the marker to see what is the next cycle, should just be the line that looks something like this `----- Cycle=... ------------------`

src/main/scala/util/util.scala
jerryz123: Thats a good point, I still think there's a way to avoid the duplication of the vecinit code (and the mostly duplicated documentation), but that can be solved later

jerryz123 (approved):

PR #258, opened 2019-05-02
Fixed doc copyright to be UCB

abejgonzalez:
@jwright6323 Mentioned that this is incorrect and should instead refer to UCB Regents.


jerryz123 (approved):

sagark:
You should also fix this line while you're at it
https://github.com/riscv-boom/riscv-boom/blob/master/LICENSE#L1

PR #260, opened 2019-05-03
Rename refactor

bkorpan:
Rewrote rename structures more concisely.


src/main/scala/exu/rename/rename-freelist.scala
```diff
@@ -1,328 +1,85 @@
...
- * @param plWidth pipeline width (dispatch group size)
- */
-class FreeListIo(
+class RenameFreeList(
+  val plWidth: Int,
   val numPregs: Int,
-  val plWidth: Int)
-(implicit p: Parameters) extends BoomBundle
+  val float: Boolean)
+  (implicit p: Parameters) extends BoomModule()(p)
```
abejgonzalez: Remember you dont need the `()(p)` anymore

bkorpan: Good catch.

PR #261, opened 2019-05-08
[rocc] Fix Rocc PNR waking

jerryz123:


jerryz123:
Received verbal approval from Ben

PR #262, opened 2019-05-10
[WIP] Bump to May rocketchip

jerryz123:


jwright6323:
We probably want https://github.com/freechipsproject/rocket-chip/pull/1943 too

colinschmidt:
We will want that but we can do it as a separate bump unless that's extremely painful.

.circleci/do-rtl-build.sh
```diff
@@ -6,17 +6,12 @@
...
+cp -r $HOME/project $HOME/bhd/generators/boom
 
 # enter the verisim directory and build the specific config
-cd $HOME/boom-template/verisim
+cd $HOME/bhd/sims/verisim
 make clean
-make CONFIG=$1 JAVA_ARGS="-Xmx2G -Xss8M"
+make SUB_PROJECT=boom CONFIG=$1 JAVA_ARGS="-Xmx2G -Xss8M"
 
-# remove generated sources to make cache smaller
```
abejgonzalez, line 20: Why was this removed?

.circleci/build-run-csmith-tests.sh
```diff
@@ -11,6 +11,6 @@ SIM=${SIM_BASE}${CONFIG}
 AMT_RUNS=$2
 
 # run csmith utility
-cd $HOME/boom-template/scripts/csmith
+cd $HOME/project/.circleci/csmith
```
abejgonzalez: Why is csmith in the `.circleci` folder, I think it should be in `util` instead since this won't reside in the `project-template` scripts section and is still useful to others.

.circleci/config.yml
```diff
@@ -272,367 +180,183 @@ jobs:
...
-                keys:
-                    - boom-template-boomconfig-{{ .Branch }}-{{ .Revision }}
-
-            - run:
-                name: Run BoomConfig assembly tests
-                command: make run-asm-tests -C ../boom-template/verisim CONFIG=BoomConfig
-                no_output_timeout: 300m
-
-    smallboomconfig-run-csmith-tests:
+                    make run-bmark-tests -C ../bhd/sims/verisim CONFIG=BoomConfig SUB_PROJECT=boom
```
abejgonzalez, line 222: Whats the reasoning to combine the tests in a single job? I doing them back to back takes a long time. If you do them separately you can potentially parallelize the runs / start up a new run when one is done.

.circleci/config.yml
```diff
@@ -272,367 +180,183 @@ jobs:
...
-                keys:
-                    - boom-template-smallboomconfig-{{ .Branch }}-{{ .Revision }}
-
-            - run:
-                name: Run SmallBoomConfig assembly tests
-                command: make run-asm-tests -C ../boom-template/verisim CONFIG=SmallBoomConfig
-                no_output_timeout: 300m
-
-    mediumboomconfig-run-csmith-tests:
+                    make run-asm-tests -C ../bhd/sims/verisim CONFIG=SmallRV32UnifiedBoomConfig SUB_PROJECT=boom
```
abejgonzalez, line 268: No bmark tests here?

src/main/scala/exu/execution-units/execution-units.scala
```diff
@@ -121,19 +121,17 @@ class ExecutionUnits(val fpu: Boolean)(implicit val p: Parameters) extends HasBo
...
-      hasIfpu        = usingFPU,
-      hasMem         = usingUnifiedMemIntIQs))
 
-    exe_units += aluExeUnit
-
-    for (w <- 0 until int_width-1) {
-      val alu_exe_unit = Module(new ALUExeUnit)
+    for (w <- 0 until int_width) {
+      def is_nth(n: Int): Boolean = w == ((n) % int_width)
+      val alu_exe_unit = Module(new ALUExeUnit(
```
abejgonzalez, line 127: Just to make sure im assuming things right. This is just to split things up "evenly" between functional units.

jerryz123: Yeah. This provides some more flexibility in allocating the functional units.

.circleci/config.yml
jerryz123: bmark tests are defaullt rv64g, not rv32.

jerryz123: bmark tests are rv64, not rv32

.circleci/config.yml
jerryz123: There is enough parallelism in 7 "Test" jobs to saturate the 4 machines we have.
Combining these significantly decreases the length of `config.yml`

.circleci/do-rtl-build.sh
jerryz123: The `path` in the cache config is `/home/user/bhd`, which does not include the `~/project` directory.

.circleci/build-run-csmith-tests.sh
jerryz123: Sure I can move it. I forgot the `util` folder existed.

.circleci/config.yml
abejgonzalez: Interesting because that didn't break the tests last time...

.circleci/config.yml
abejgonzalez: I think having a long `config.yml` isn't a bad thing. I would prefer to have these as separate jobs. Not only would it make it more sense IF we get more parallelism in the future but it makes it cleaner to read.

.circleci/config.yml
jerryz123: We never ran bmark-tests for SmallRV32Unified
https://github.com/riscv-boom/riscv-boom/blob/317b13756bdee864a8f84cad6258633273252201/.circleci/config.yml#L686

.circleci/config.yml
jerryz123: Having long redundant files also makes maintenance difficult

If you want you can PR this back later.

abejgonzalez: Fine. We can do this for now.

.circleci/config.yml
abejgonzalez: My bad I misremembered

abejgonzalez (approved):

PR #268, opened 2019-05-20
Fake chicken-bit CSR

jerryz123:
Wait for #262 to merge.
Needed for compatibility with rocket-chip bootrom, which writes to the chicken-bit CSRs.

### DIFF
--- a/src/main/scala/common/parameters.scala
+++ b/src/main/scala/common/parameters.scala
@@ -12,6 +12,7 @@ import chisel3.util._
 
 import freechips.rocketchip.rocket._
 import freechips.rocketchip.tile._
+import freechips.rocketchip.util._
 import freechips.rocketchip.config.{Parameters, Field}
 
 import boom.ifu._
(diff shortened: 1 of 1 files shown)


src/main/scala/common/parameters.scala
```diff
@@ -89,13 +90,31 @@ case class BoomCoreParams(
...
+  val nPMPs: Int = 0 // TODO Fix this!!!!!
+
+  override def customCSRs(implicit p: Parameters) = new BoomCustomCSRs
+}
+
+/**
+  * Defines custom BOOM CSRs
+  */
+class BoomCustomCSRs(implicit p: Parameters) extends freechips.rocketchip.tile.CustomCSRs
+    with HasBoomCoreParameters {
```
abejgonzalez: Can this be 2 spaces for indentation? I know it looks kinda weird with the indentation on the next line but I would just put a space afterward.

abejgonzalez: @jerryz123 This is good to merge pending spacing fixes

abejgonzalez (approved):

PR #271, opened 2019-05-24
Heterogenous Multi-core Support

abejgonzalez:
- Reorganize configs a bit and rename map variables
- Add new BOOM + Rocket subsystem
- Make default subsystem be BOOM + Rocket

Decisions to be made...
- Delete the previous BOOM only subsystem?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -219,8 +219,8 @@ jobs:
             - run:
                 name: Run BoomConfig riscv tests
                 command: |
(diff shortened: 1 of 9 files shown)


jerryz123 (commented):
So it looks like we have to update the BHD default `SUB_PROJECT=boom` flags to use the ExampleBoomAndRocketSystem as well?

abejgonzalez:
@jerryz123 I addressed this in the https://github.com/ucb-bar/project-template/pull/92. However, I do think the Ci should stay the same (technically, I would want all the BHD parameters indicated in our CI but that gets long so this'll do for now)

jerryz123 (approved):

PR #272, opened 2019-05-25
add support for boom being first hart

abejgonzalez:


jerryz123:
BOOM should always be first >:)

abejgonzalez:
But BOOM is out of order right?

jerryz123 (approved):

PR #277, opened 2019-06-05
BOOM Crossings

abejgonzalez:
Similar to how Rocket supports different clock crossings on the tile-level, now duplicate `RocketCrossingKey` into the `BoomCrossingKey`. This makes it clearer to split up the two different sets of tiles and adds its own set of mixins.

Functionally, this is no different from the `RocketCrossingKey`, it just makes it easier to understand in the code.

Finally, there are minor changes in naming the tile and comments.

### DIFF
--- a/src/main/scala/system/BoomSubsystem.scala
+++ b/src/main/scala/system/BoomSubsystem.scala
@@ -20,7 +20,10 @@ import freechips.rocketchip.tile._
 import freechips.rocketchip.subsystem._
 import freechips.rocketchip.amba.axi4._
 
+import boom.common.{BoomTile}
+
 case object BoomTilesKey extends Field[Seq[boom.common.BoomTileParams]](Nil)
+case object BoomCrossingKey extends Field[Seq[RocketCrossingParams]](List(RocketCrossingParams()))
 
 trait HasBoomTiles extends HasTiles
   with CanHavePeripheryPLIC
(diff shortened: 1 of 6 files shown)


src/main/scala/system/BoomSubsystem.scala
```diff
@@ -31,21 +34,21 @@ trait HasBoomTiles extends HasTiles
   val module: HasBoomTilesModuleImp
 
   protected val boomTileParams = p(BoomTilesKey)
-  private val crossings = perTileOrGlobalSetting(p(RocketCrossingKey), boomTileParams.size)
+  // crossing can either be per tile or global (aka only 1 crossing specified)
```
colinschmidt, line 37: Global here means for all booms not all tiles right.

abejgonzalez: In this case, for all BOOM tiles. Specifically, if there is only 1 item in the `BoomCrossingKey` `Seq`, that applies to all tiles (all the ones that use `crossings`). However, if there is more than 1 item in the `Seq` then there must be `N` of them (corresponding to the amt of tiles.

Does that make more sense?

abejgonzalez:
@colinschmidt Any other thoughts?

colinschmidt (approved):
LGTM.

jerryz123 (approved):

bkorpan (approved):

PR #278, opened 2019-06-06
Fix Interrupt/Hart Mapping and BOOM Hart 0 Default

abejgonzalez:
Fixes a bug where assigning `hartId` of 0 to BOOM would not match with that the interrupts would be assigned in the PLIC and CLINT. This was due to the ordering of the `connectInterrupts` function.

Additionally, have the `WithRenumberHarts` have BOOM harts be ordered before Rocket harts.

### DIFF
--- a/src/main/scala/system/BoomAndRocketSubsystem.scala
+++ b/src/main/scala/system/BoomAndRocketSubsystem.scala
@@ -48,15 +48,11 @@ trait HasBoomAndRocketTiles extends HasTiles
 
     connectMasterPortsToSBus(rocket, crossing)
     connectSlavePortsToCBus(rocket, crossing)
-    connectInterrupts(rocket, Some(debug), clintOpt, plicOpt)
+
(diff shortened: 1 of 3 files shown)


abejgonzalez:
Im pretty sure this is a bug with the `RocketSubsystem`. I assume they don't allow cores to arbitrarily change their `hartId` which will break the CLINT interrupt mapping.

Checking as we speak.

abejgonzalez:
I can confirm that this issue also does present itself in Rocket.

abejgonzalez:
Since this will apply to things in `project-template`... any thoughts @colinschmidt (are there some rocket-chip people I should add?)

colinschmidt (approved):
LGTM. We could probably upstream the sort by hartID thing but it wouldn't help project-template anyways so its probably not worth it. Unless you have a simple config that renumbers harts and breaks rocket-chip.

bkorpan (approved):

jerryz123 (approved):

PR #279, opened 2019-06-09
Fetch Buffer Refactor

bkorpan:
Refactored fetch buffer enqueue/dequeue logic to improve QoR for wide frontends.

Instead of compaction occurring before writing into the RAM (register array), compaction occurs by controlling the write indices. This means that enqueuing instructions to not need to pass through a large 'muxing pyramid': they stay in their respective 'lanes'. The write indices use a one-hot representation.

This style of fetch buffer required the output register be removed, somewhat lengthening the subsequent (decode/rename) stage. To help with this, the dequeue logic was made to use a single read port, requiring the fetch buffer size is a multiple of the decode width. This also prevents N < decodeWidth instructions from being dequeued: is this a problem?

### DIFF
--- a/src/main/scala/ifu/fetch-buffer.scala
+++ b/src/main/scala/ifu/fetch-buffer.scala
@@ -23,7 +23,7 @@ import chisel3.core.{DontCare}
 import freechips.rocketchip.config.{Parameters}
 
 import boom.common._
-import boom.util.{BoolToChar}
+import boom.util.{BoolToChar, WrapInc}
 
 /**
  * Bundle that is made up of converted MicroOps from the Fetch Bundle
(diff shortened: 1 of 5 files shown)


jerryz123:
Since it looks like the interface to the fetch-buffer is the same, we should consider adding a flag `enableNewFetchBuffer`, and then use your new module when that flag is set.

By default we can leave this flag unset, and in the BEAGLE config we would override it.
This lets us get this into master without having to worry about QoR or performance or correctedness yet.

When we finish validating the new design, we can deprecate the flag, and delete the old module, and replace it.

src/main/scala/ifu/fetch-buffer.scala
```diff
@@ -52,40 +52,44 @@ class FetchBuffer(num_entries: Int)(implicit p: Parameters) extends BoomModule
...
+  def rotateLeft(in: UInt, k: Int) = {
+    val n = in.getWidth
+    Cat(in(n-k-1,0), in(n-1, n-k))
+  }
+
+  val might_hit_head = (1 until fetchWidth).map(k => VecInit(rotateLeft(tail, k).asBools.zipWithIndex.filter
+    {case (e,i) => i % coreWidth == 0}.map {case (e,i) => e}).asUInt).map(tail => head & tail).reduce(_|_).orR
+  val at_head = (VecInit(tail.asBools.zipWithIndex.filter {case (e,i) => i % coreWidth == 0}
+    .map {case (e,i) => e}).asUInt & head).orR
+  val do_enq = !(at_head & maybe_full || might_hit_head)
```
abejgonzalez: `at_head` and `maybe_full` are both bools right. So to make this a bit clearer, I would do `&&` to indicate that they are bools).

bkorpan: Nice catch.

jerryz123:
Merging so BEAGLE can use boom/master.
We'll validate the new fetch-buffer later.

jerryz123 (approved):

PR #281, opened 2019-06-09
Common refactor

bkorpan:
Refactored core configs in common/configs.scala

Replaced DefaultBoomConfig with BaseBoomConfig. BaseBoomConfig is not a valid core configuration, a core config must be mixed in.

4 core configs now: Small, Medium, Large and Mega which feature 1, 2, 3 and 4 -wide pipelines.

Style refactor of common package (spacing, braces, etc.)
This includes some renaming of signals & parameters used throughout the codebase.


src/main/scala/system/Configs.scala
```diff
@@ -163,40 +173,42 @@ class SmallRV32UnifiedBoomConfig extends Config(
...
+class MediumBoomAndRocketConfig extends Config(
   // final param setup
   new WithRenumberHarts ++
   new WithoutTLMonitors ++
   // boom param setup
   new WithRVC ++
-  new WithSmallBooms ++
-  new DefaultBoomConfig ++
+  new BaseBoomConfig ++
+  new WithMediumBooms ++
```
jerryz123, line 211: Is this in the right order?

bkorpan: > 
> 
> Is this in the right order?

No, but it doesn't actually matter right now. It will if we add some parameters to BaseBoomConfig which overlap with the core configs.

jerryz123:
Is LargeBoomConfig appropriately sized for BEAGLE?

bkorpan:
> 
> 
> Is LargeBoomConfig appropriately sized for BEAGLE?

Yeah, I think so.

src/main/scala/common/configs.scala
```diff
@@ -23,49 +23,23 @@ import boom.lsu._
...
-   case SystemBusKey => up(SystemBusKey, site).copy(beatBytes = 16)
-
-   // Make sure there are enough hart bits to support multiple cores
-   case MaxHartIdBits => log2Up(site(BoomTilesKey).size)
+class BaseBoomConfig extends Config((site, here, up) => {
+  // Top-Level
+  case XLen => 64
+
+  // Use this boot ROM for SimDTM.
+  case BootROMParams => BootROMParams(contentFileName = s"./bootrom/bootrom.rv${site(XLen)}.img")
```
abejgonzalez: Can this move to a another mixin? Technically, the bootrom is not related to the core parameters and can change depending on what the configuration needs. Also... removing this makes it fall in line with what `project-template` does

src/main/scala/system/Configs.scala
```diff
@@ -163,40 +173,42 @@ class SmallRV32UnifiedBoomConfig extends Config(
...
 
-class DualCoreBoomAndDualRocketConfig extends Config(
+class DualCoreMediumBoomAndDualRocketConfig extends Config(
   // final param setup
   new WithRenumberHarts ++
   new WithoutTLMonitors ++
   // boom param setup (applies to all boom cores)
   new WithRVC ++
-  new DefaultBoomConfig ++
+  new BaseBoomConfig ++
```
abejgonzalez: These should be swapped?

src/main/scala/lsu/lsu.scala
```diff
@@ -170,70 +170,70 @@ class LoadStoreUnit(pl_width: Int)(implicit p: Parameters,
    val io = IO(new LoadStoreUnitIO(pl_width))
 
    // Load-Address Queue
-   val laq_addr_val       = Reg(Vec(NUM_LDQ_ENTRIES, Bool()))
```
jerryz123: Can you revert all the changes to LSU.scala, and add both NUM_LDQ_ENTRIES and numLdqEntries to the set of variables? Changing the naming here makes rebasing the LSU refactor painful. 
I'll change the naming in that PR.

jerryz123 (changes requested):

src/main/scala/system/Configs.scala
bkorpan: Good catch, I could have sworn I already fixed that.

src/main/scala/system/Configs.scala
```diff
@@ -38,44 +38,45 @@ import boom.common._
...
   new WithNBoomCores(1) ++
   new WithoutTLMonitors ++
   new freechips.rocketchip.system.BaseConfig)
 
 class MediumBoomConfig extends Config(
   new WithRVC ++
   new WithMediumBooms ++
-  new DefaultBoomConfig ++
+  new BaseBoomConfig ++
```
abejgonzalez, line 65: IMO. Having the `BaseBoomConfig` after the `WithNBoomCores` seems weird. Ontop of that, having `WithNBoomCores` specifying parameters that can override prior mixins, I feel like `WithNBoomCores` should just take the prior tile params specified and duplicate the tiles. i.e.

if (tile param prev. specified) -> duplicate those params N times
else -> create N generic cores

src/main/scala/common/parameters.scala
```diff
@@ -24,71 +24,71 @@ import boom.lsu._
...
+    IssueParams(issueWidth=1, numEntries=16, iqType=IQT_FP.litValue , dispatchWidth=1)),
+  numLdqEntries: Int = 16,
+  numStqEntries: Int = 16,
+  numIntPhysRegisters: Int = 96,
+  numFpPhysRegisters: Int = 64,
+  enableCustomRf: Boolean = false,
+  enableCustomRfModel: Boolean = true,
+  maxBrCount: Int = 4,
+  fetchBufferSz: Int = 8,
+  useNewFetchBuffer: Boolean = true,
```
abejgonzalez: If this is enabled then we should delete the old one.

bkorpan: Another good catch, it should be disabled by default.

src/main/scala/common/parameters.scala
```diff
@@ -123,175 +123,166 @@ class BoomCustomCSRs(implicit p: Parameters) extends freechips.rocketchip.tile.C
...
+  //************************************
+  // Implicitly calculated constants
+  val numRobRows      = numRobEntries/coreWidth
+  val robAddrSz       = log2Ceil(numRobEntries)
+  // the f-registers are mapped into the space above the x-registers
+  val logicalRegCount = if (usingFPU) 64 else 32
+  val lregSz          = log2Ceil(logicalRegCount)
+  val ipregSz         = log2Ceil(numIntPhysRegs)
+  val fpregSz         = log2Ceil(numFpPhysRegs)
+  val pregSz          = ipregSz max fpregSz
```
abejgonzalez: Instead of overriding this in particular places, I would instead just rename this `maxPregSz` or something similar to that. Or rename the inner variable in the submodules to be something like `localPregSz`/`freelistPregSz`

src/main/scala/system/Configs.scala
abejgonzalez: Now that I think of it... this might be hard to do without it getting overly confusing. I think for now having the `Base` config be after is fine until I think of a better solution.

src/main/scala/common/parameters.scala
bkorpan: `maxPregSz` works

src/main/scala/system/Configs.scala
```diff
@@ -163,40 +186,45 @@ class SmallRV32UnifiedBoomConfig extends Config(
...
+  new WithMediumBooms ++
+  new BaseBoomConfig ++
   // create boom tile
   new WithNBoomCores(1) ++
   // create rocket tile
   new freechips.rocketchip.subsystem.WithNBigCores(1) ++
   new freechips.rocketchip.system.BaseConfig)
 
-class DualCoreBoomAndDualRocketConfig extends Config(
+class DualCoreMediumBoomAndDualRocketConfig extends Config(
```
abejgonzalez: While we are at it, lets get rid of the "core" in the name. Or switch to "DualCoreMediumBoomAndDualCoreRocketConfig"

abejgonzalez (approved):
LGTM. I think this is ready to be merged in. We have time to adjust for the BEAGLE tapeout (plus this doesn't involve any RTL changes)

jerryz123 (approved):

PR #282, opened 2019-06-10
speed up ci by adding fast (removes verbose print)

abejgonzalez:
My reasoning for defaulting to fast is so that verbose does not print. However, you still do get an error on asserts. It is just up to the PR owner to rerun that test with verbose.


bkorpan (approved):

jerryz123:
The failing test was resolved in #284

jerryz123 (approved):

PR #286, opened 2019-06-10
Have CSmith CI use TSI Bringup

abejgonzalez:
Must wait until #285 is merged in first. This is a split of #283 where CSmith now uses the TSI bringup given by `project-template`


abejgonzalez:
Closing as this needs to be addressed with more memory in CI

PR #289, opened 2019-06-11
ROB Head & Tail Fix

bkorpan:
Fixed behavior of ROB head and tail lower order bits for pipeline widths > 3.


jerryz123:
Let's delay merging these until the stq issue is resolved.

abejgonzalez (approved):

PR #294, opened 2019-06-12
add issue and pull request templates based on rc/chisel/firrtl

abejgonzalez:
**Related issue**: none

**Impact**: no functional change

**Release Notes**
This will make it easier for changes to be tracked across multiple PRs that are merged. This is a proposal for how it should be (matches other busy projects) but it can change. Thoughts?

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,15 +4,15 @@
 <!-- if the issue is NOT a question, fill out the following sections -->
 
 <!-- choose one -->
-**Impact**: no functional change | API addition (no impact on existing code) | API modification | unknown
+**Impact**: no rtl change | rtl refactoring | new rtl | unknown
 
 <!-- choose one -->
 **Development Phase**: request | proposal
 
 **Other information**
- <!-- include detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. Stack Overflow, gitter, etc -->
+<!-- include detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context -->
 
-<!-- if the current behavior is a bug, please provide the steps to reproduce the problem -->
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
+<!-- choose all that apply -->
+**Type of issue**: bug report | feature request | question | other enhancement
+
+<!-- if the issue is NOT a question, fill out the following sections -->
+
+<!-- choose one -->
+**Impact**: no functional change | API addition (no impact on existing code) | API modification | unknown
```
jerryz123: Lets think of a categorization that makes more sense for RTL.
The categories I see are
- No RTL change
- RTL refactoring (equivalent functionality)
- New RTL

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+
+<!-- choose one -->
+**Development Phase**: request | proposal
+
+**Other information**
+ <!-- include detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. Stack Overflow, gitter, etc -->
```
jerryz123: We don't have a stack overflow or gitter.

abejgonzalez: So this was copied but I dont think it actually hurts us to have this... its just saying... give us context.

.github/ISSUE_TEMPLATE.md
abejgonzalez: That sounds reasonable. I think we can re-add the API categories if we eventually move to a reusability model (e.g. instead of having everything so tightly coupled with RC parameterization).

bkorpan (approved):

jerryz123 (approved):

PR #297, opened 2019-06-14
[lsu] Don't forward out loads to uncacheable regions

jerryz123:


donggyukim:
I just found out some related issue when running `blkdev.riscv` from project template. In this test, MMIOs were out of order... Do you think this PR will fix it?

jerryz123:
Yes.

abejgonzalez (approved):

abejgonzalez:
@donggyukim This solved an issue in the bootrom where the CLINT writes (MMIO) would be forwarded to a corresponding LD. What do you mean OOO?

PR #298, opened 2019-06-15
Fixed ROB fullness tracking w.r.t. mispredicts

bkorpan:
Make sure maybe_full is true after a mispredict.

### DIFF
--- a/src/main/scala/exu/rob.scala
+++ b/src/main/scala/exu/rob.scala
@@ -795,7 +795,7 @@ class Rob(
   // I.E. at least one entry will be empty when in a steady state of dispatching and committing a row each cycle.
   // TODO should we add an extra 'parity bit' onto the ROB pointers to simplify this logic?
 
-  maybe_full := !rob_deq && (rob_enq || maybe_full)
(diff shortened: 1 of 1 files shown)


src/main/scala/exu/rob.scala
```diff
@@ -795,7 +795,7 @@ class Rob(
   // I.E. at least one entry will be empty when in a steady state of dispatching and committing a row each cycle.
   // TODO should we add an extra 'parity bit' onto the ROB pointers to simplify this logic?
 
-  maybe_full := !rob_deq && (rob_enq || maybe_full)
+  maybe_full := !rob_deq && (rob_enq || maybe_full) || io.brinfo.mispredict
```
jerryz123, line 798: Shouldn't this be `io.brinfo.valid && io.brinfo.mispredict`

jerryz123 (changes requested):

src/main/scala/exu/rob.scala
bkorpan: io.brinfo.mispredict can't go high without a valid brinfo packet.

jerryz123 (approved):

PR #299, opened 2019-06-16
Core cleanup

bkorpan:
Removed large spans of commented out code as part of of an ongoing effort to reduce the LoC of core.scala


abejgonzalez (approved):
LGTM

PR #300, opened 2019-06-19
cleanup prints | have multicore configs show properly in collateral

abejgonzalez:
**Type of change**: bug fix + other enhancement

**Impact**: no rtl change

**Release Notes**
This adds a new function called `AddToStringPrefix` that makes it so that the prefix for all the BOOM prints is in one location. Additionally, there is a couple of new print headers and spaces to make things a bit clearer. Finally, the `*.core.config` file that is generated is made in the subsystem level so that it can include all tiles in the system rather than just the first.

### DIFF
--- a/src/main/scala/ifu/frontend.scala
+++ b/src/main/scala/ifu/frontend.scala
@@ -32,7 +32,7 @@ import boom.bpu._
 import boom.common._
 import boom.exu.{BranchUnitResp, CommitExceptionSignals}
 import boom.lsu.{CanHaveBoomPTW, CanHaveBoomPTWModule}
-import boom.util.{AddToStringPrefix}
+import boom.util.{BoomCoreStringPrefix}
 
 /**
  * Parameters to manage a L1 Banked ICache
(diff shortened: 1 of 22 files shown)


src/main/scala/util/util.scala
```diff
@@ -572,3 +574,17 @@ object FPRegToChars
     multiVec(fpreg)
   }
 }
+
+object AddToStringPrefix
```
bkorpan: Would a more specific name like "BuildCoreInfoString" make sense?

src/main/scala/ifu/frontend.scala
```diff
@@ -320,7 +321,12 @@ class BoomFrontendModule(outer: BoomFrontend) extends LazyModuleImp(outer)
   def ccover(cond: Bool, label: String, desc: String)(implicit sourceInfo: SourceInfo) =
     cover(cond, s"FRONTEND_$label", "Rocket;;" + desc)
 
-  override def toString: String = bpdpipeline.toString + "\n" + icache.toString
+  override def toString: String =
+    (AddToStringPrefix("====Overall Frontend Params====")
+    + "\n"
+    + bpdpipeline.toString
+    + "\n"
+    + icache.toString)
```
bkorpan, line 327: Not a fan of having "\n" on their own line.

jerryz123: I agree

src/main/scala/util/util.scala
abejgonzalez: I agree the function could change. Maybe `BoomCoreStringPrefix` also? Do you like `prefix` in the name?

src/main/scala/ifu/frontend.scala
abejgonzalez: I can change that.

src/main/scala/util/util.scala
bkorpan: Sure

bkorpan (approved):

PR #303, opened 2019-06-22
Fetch Buffer Partial Dequeue

bkorpan:
Let the new fetch buffer perform partial dequeues.


jerryz123:
Do the rob changes break with the old fetchbuffer

bkorpan:
> Do the rob changes break with the old fetchbuffer

No reason they should.

jerryz123:
The modifications to the rob pointer lsb tracking are compatible with both?

bkorpan:
> The modifications to the rob pointer lsb tracking are compatible with both?

Yeah, they should be. I haven't tested extensively.

abejgonzalez (approved):

PR #304, opened 2019-06-22
Rename Refactor

bkorpan:

**Impact**: rtl refactoring

**Release Notes**
Rewrote rename structures more concisely and refactored the decode/rename/dispatch pipeline control logic.


src/main/scala/exu/core.scala
```diff
@@ -166,6 +173,10 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
   val brunit_idx = exe_units.br_unit_idx
   br_unit <> exe_units.br_unit_io
 
+  val flush = br_unit.brinfo.mispredict ||
```
ccelio: Stylistically, I find this a little confusing that the LHS is `flush` while the RHS has `rob.io.flush.valid` as 1 of 3 possibilities. Perhaps `flush_ifu` to clarify that it is covering the complete _nuking_ of the IFU? Naming is tough. :(

src/main/scala/exu/core.scala
```diff
@@ -149,13 +149,20 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
   //***********************************
   // Pipeline State Registers and Wires
 
-  // Instruction Decode Stage
+  // Decode/Rename1 Stage
   val dec_valids    = Wire(Vec(coreWidth, Bool()))  // are the decoded instruction valid? It may be held up though.
   val dec_uops      = Wire(Vec(coreWidth, new MicroOp()))
-  val dec_will_fire = Wire(Vec(coreWidth, Bool()))  // can the instruction fire beyond decode?
+  val dec_proceed   = Wire(Vec(coreWidth, Bool()))  // can the instruction fire beyond decode?
```
ccelio: There is definitely room for clarifying the differences between "valid, mask, proceed, can_fire, will_fire, fire, etc." (`can_fire` and `will_fire` are Bluespec-isms). In this case, I'm wondering if `dec_fire` is better than `dec_proceed` (and consistent with the `dis_fire` and `ren2_fire`, since Chisel's decoupled bundles have `ready`, `valid`, and `fire()`? Or is `proceed` better since there are situations where it is necessary (like in the LSU) to distinguish between `can_fire` and `will_fire`? I wonder what Hwacha's code style does here?

src/main/scala/exu/rename/rename-busytable.scala
```diff
@@ -1,211 +1,64 @@
-//******************************************************************************
```
ccelio: Don't delete copyright header.

src/main/scala/exu/rename/rename-freelist.scala
```diff
@@ -1,328 +1,85 @@
...
-      enq_mask(w) := 1.U << io.enq_pregs(w)
-    } .elsewhen (io.rollback_wens(w)) {
-      enq_mask(w) := 1.U << io.rollback_pdsts(w)
-    }
-  }
+  // Allocations seen by branches in each pipeline slot.
+  val alloc_masks = (preg_sels zip io.reqs).scanRight(0.U(numPregs.W)) {case ((preg, req), mask) => Mux(req, mask | preg, mask)}
 
+  // Pregs returned by the ROB via commit or rollback.
+  val ret_valids = Mux(io.rollback, io.rbk_valids, io.com_valids)
```
ccelio: Are rbk_valids and com_valids mutually exclusive? This mux feels weird to me. I think an assertion that guards mutual exclusivity would be nice and then some of this logic may simplify?

src/main/scala/exu/rename/rename-freelist.scala
```diff
@@ -1,328 +1,85 @@
...
-    // set other branch allocation_lists to zero where allocation_list(j) == 1...
-    for (i <- 0 until maxBrCount) {
-      allocation_lists(i) := allocation_lists(i) & ~allocation_list
-    }
+  when (io.brinfo.mispredict) {
+    // Recover pregs allocated past a mispredicted branch.
+    free_list := free_list | br_alloc_lists(io.brinfo.tag) | ret_mask
+  } .otherwise {
+    // Update the free list.
+    free_list := free_list & ~alloc_masks(0) | ret_mask
```
ccelio, line 86: I think this code should have a bunch of asserts thrown in to guard against invalid invariants. For example, verify that you are never freeing a register that is already freed.

src/main/scala/exu/rename/rename-maptable.scala
```diff
@@ -1,323 +1,96 @@
...
-    }
-    rs3_cases ++= Array((io.ren_uops(w).frs3_en, map_table_prs3(w)))
-
-    // Set outputs.
-    io.values(w).prs1       := MuxCase(io.ren_uops(w).lrs1, rs1_cases)
-    io.values(w).prs2       := MuxCase(io.ren_uops(w).lrs2, rs2_cases)
-    if (rtype == RT_FLT.litValue)
-      io.values(w).prs3     := MuxCase(io.ren_uops(w).lrs3, rs3_cases)
-    io.values(w).stale_pdst := MuxCase(map_table_io(io.ren_uops(w).ldst).element, stale_cases)
+  when (io.brinfo.mispredict) {
```
ccelio, line 93: What happens if `mispredict` and `rollback` are both true?

ccelio (approved):
At a high-level, my only real request is that you think hard about adding asserts to the rename logic (esp. the freelist and maptable). I was too lazy and complacent to do so before, but I think now is a good time to do so while the logic is fresh in your mind.

src/main/scala/exu/core.scala
bkorpan: I thought it made more sense to think of `fire` as a property of the pipeline register - firing means it is latching new values. Proceeding is a property of the things in the pipeline slots - they proceed if they move to the next stage and their valid bit is cleared (or replaced if that stage also fires).

src/main/scala/exu/rename/rename-freelist.scala
bkorpan: Good idea. I had also thought about an assert that protects against duplicate mappings in the Maptable, and this refactor should make that much easier to write.

src/main/scala/exu/rename/rename-freelist.scala
bkorpan: You're right, they should be mutually exclusive. It might be better to have a single valid vector in the ROB commit interface.

src/main/scala/exu/rename/rename-maptable.scala
bkorpan: That shouldn't be able to happen.

src/main/scala/common/parameters.scala
```diff
@@ -246,6 +246,7 @@ trait HasBoomCoreParameters extends freechips.rocketchip.tile.HasCoreParameters
   //************************************
   // Extra Knobs and Features
   val enableCommitMapTable = boomParams.enableCommitMapTable
+  require(!enableCommitMapTable)
```
jerryz123: I recommend deleting all traces of the `commitMapTable`. Its definitely very broken, and I don't think we will ever re-enable it again.

src/main/scala/exu/core.scala
```diff
@@ -385,43 +390,40 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
 
+  // Decode/Rename1 pipeline logic
+
+  val dec_prior_slot_valid = dec_valids.scanLeft(false.B) ((s,v) => s || v)
+  val dec_prior_slot_unique = dec_uops.scanLeft(false.B) ((s,u) => s || u.is_unique)
+  val wait_for_empty_pipeline = (0 until coreWidth).map(w => dec_uops(w).is_unique &&
+                                  (!rob.io.empty || !lsu.io.lsu_fencei_rdy || dec_prior_slot_valid(w)))
+  val wait_for_rocc = (0 until coreWidth).map(w =>
+                        (dec_uops(w).is_fence || dec_uops(w).is_fencei) && (io.rocc.busy || rocc_shim_busy))
+
```
jerryz123, line 405: stylistic thing, but these make more sense as functions accepting a index, rather than vectors. 
`def wait_for_rocc(i) = (dec_uops(i).is_fence || dec_uops(w).is_fencei) && (io.rocc.busy || rocc_shim_busy)`

Seems slightly more concise, IMO

src/main/scala/exu/core.scala
jerryz123: Personally I like the `can_fire`, `will_fire` naming.

src/main/scala/exu/core.scala
```diff
@@ -432,63 +434,136 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
+                      dis_valids(w) &&
+                      ( !dispatcher.io.ren_uops(w).ready
+                      || flush))
+
+  val dis_stalls = dis_hazards.scanLeft(false.B) ((s,h) => s || h).takeRight(coreWidth)
+  dis_proceed := dis_valids zip dis_stalls map {case (v,s) => v && !s}
+  dis_fire := !dis_stalls.last
+
+  // Decode pipeline logic either sees a single "fire/ready" signal,
+  //   or we combine the dec/dis hazards when using a single-stage rename pipeline.
```
jerryz123: I think you should just deprecated single-stage rename in this PR as well.

jerryz123 (approved):

src/main/scala/exu/core.scala
bkorpan: I can flip the roles of `proceed` and `fire` back around if that's the consensus.

src/main/scala/common/parameters.scala
bkorpan: I might hold off on this, it doesn't seem exorbitantly difficult to fix.

src/main/scala/exu/rob.scala
```diff
@@ -990,12 +987,12 @@ class Rob(
...
 
-  override def toString: String = BoomCoreStringPrefix(
-    "==ROB==",
-    "Machine Width      : " + coreWidth,
-    "Rob Entries        : " + numRobEntries,
-    "Rob Rows           : " + numRobRows,
-    "Rob Row size       : " + log2Ceil(numRobRows),
-    "log2Ceil(coreWidth): " + log2Ceil(coreWidth),
-    "FPU FFlag Ports    : " + numFpuPorts)
+  override def toString: String =
```
bkorpan: Rebasing mistake here, I'll fix this soon.

src/main/scala/exu/core.scala
bkorpan: I can do that.

src/main/scala/exu/core.scala
bkorpan: That is a bit more elegant, but then you don't get convenient waveform visibility of `(dec_uops(i).is_fence || dec_uops(w).is_fencei) && (io.rocc.busy || rocc_shim_busy)`.

jerryz123: That's true, I hadn't considered that. I think the waveform visibility is worth it

src/main/scala/exu/rename/rename-maptable.scala
ccelio: Then you should add an assert to enforce that that can never happen. I try to think of asserts as "if someone was to come along and add formal; what things do I need to codify for them?"

src/main/scala/common/parameters.scala
ccelio: I'd actually go in the other direction and recommend fixing and using the commitMapTable. But maybe not for next tapeout.

src/main/scala/exu/rename/rename-maptable.scala
bkorpan: Agreed.

src/main/scala/exu/rename/rename-freelist.scala
ccelio: I love that idea.

src/main/scala/exu/core.scala
```diff
@@ -176,7 +176,8 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
                          rob.io.flush.valid || // i.e. 'flush in-order part of the pipeline'
                          io.ifu.sfence_take_pc
 
-  assert (!(br_unit.brinfo.mispredict && rob.io.commit.rollback), "Can't have a mispredict during rollback.")
+  assert (!(br_unit.brinfo.mispredict && rob.io.commit.rollback) || rob.io.flush.valid,
+    "Can only have a mispredict during the first cycle of rollback.")
```
ccelio: I think this commit can now be rolled back since you moved rollback to start a cycle later.

bkorpan: Sounds good.

bkorpan:
This should be ready to go now. I'll do another check that it boots Fedora after this big round of tweaks.

bkorpan:
Boots Fedora fine with 2, 3, and even 4 -wide pipelines.

src/main/scala/exu/core.scala
```diff
@@ -385,43 +391,40 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
+  val wait_for_empty_pipeline = (0 until coreWidth).map(w => dec_uops(w).is_unique &&
+                                  (!rob.io.empty || !lsu.io.lsu_fencei_rdy || dec_prior_slot_valid(w)))
+  val wait_for_rocc = (0 until coreWidth).map(w =>
+                        (dec_uops(w).is_fence || dec_uops(w).is_fencei) && (io.rocc.busy || rocc_shim_busy))
+
+  val dec_hazards = (0 until coreWidth).map(w =>
+                      dec_valids(w) &&
+                      (  !rob.io.ready
+                      || !dis_ready
+                      ||  branch_mask_full(w)
```
abejgonzalez: Extra space

src/main/scala/exu/core.scala
```diff
@@ -432,63 +435,130 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
   for (w <- 0 until coreWidth) {
     // note: this assumes uops haven't been shifted - there's a 1:1 match between PC's LSBs and "w" here
     // (thus the LSB of the rob_idx gives part of the PC)
     if (coreWidth == 1) {
-       dec_uops(w).rob_idx := rob.io.rob_tail_idx
+      dec_uops(w).rob_idx := rob.io.rob_tail_idx
     } else {
       dec_uops(w).rob_idx := Cat(rob.io.rob_tail_idx >> log2Ceil(coreWidth).U,
-                                 w.U(log2Ceil(coreWidth).W))
+                                w.U(log2Ceil(coreWidth).W))
```
abejgonzalez: Shouldn't this be moved over one more space?

bkorpan: Yup.

abejgonzalez (approved):

PR #308, opened 2019-06-24
4-stage fetch

bkorpan:

**Impact**: rtl refactoring

**Release Notes**
Added another stage to instruction fetch. The fetch bundle path is split by an f4 register prior to enqueueing to the fetch buffer. The ftq still enqueues fetch targets in f3, as it is not the endpoint of any long paths.

### DIFF
--- a/src/main/scala/ifu/fetch-control-unit.scala
+++ b/src/main/scala/ifu/fetch-control-unit.scala
@@ -144,10 +144,10 @@ class FetchControlUnit(implicit p: Parameters) extends BoomModule
 
   // F3 is invalidated by redirects.
   val f3_valid_masked = f3_valid && !r_f4_req.valid
-  val f4_fire = f3_valid_masked && f4_ready
+  val f3_fire = f3_valid_masked && f4_ready
 
   // F4 Instruction path.
-  val r_f4_fetch_bundle = RegEnable(f3_fetch_bundle, f4_fire)
+  val r_f4_fetch_bundle = RegEnable(f3_fetch_bundle, f3_fire)
 
   io.f3_stall := !f4_ready
   io.f3_clear := clear_f3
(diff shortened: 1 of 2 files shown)


src/main/scala/ifu/fetch-control-unit.scala
```diff
@@ -122,21 +122,30 @@ class FetchControlUnit(implicit p: Parameters) extends BoomModule
...
+  val f3_req          = Wire(Valid(new PCReq()))
+  val f3_fetch_bundle = Wire(new FetchBundle)
+  val f3_valid        = q_f3_imemresp.io.deq.valid
 
   dontTouch(f3_fetch_bundle)
 
   val r_f4_valid = RegInit(false.B)
+
+  // Can the F3 stage proceed?
+  val f4_ready = (!r_f4_valid || fb.io.enq.ready) && ftq.io.enq.ready
```
jerryz123: Could this be simpler with a 1-deep queue.

I also think it may be a good idea to make this configurable (for now).
We need to make a graph showing performance impacts of all the random stuff that's changed.

bkorpan: I had thought about doing that, I can look at it. I held off because these options tend to be messy code-wise, and it takes longer to verify both settings (and we inevitably end up wanting to deprecate them).

jerryz123: Yeah I agree it's messy,and hopefully we would depreciate it soon.

It's just this may have a noticeable performance impact, and it would be good to document that with a good comparison.

bkorpan: Yeah, makes sense.

jerryz123:
Also, depending on what's default, some of the documentation should be updated

bkorpan:
> Also, depending on what's default, some of the documentation should be updated

Sure.

abejgonzalez (approved):
LGTM. However, we should add a comment saying that 4 stage is really the only supported thing here. Or if we are going to not use 3 stage, just remove support for it completely.

PR #313, opened 2019-07-01
Back nPMPs

pentin-as:
I did not understand why PMP were excluded. If enabled, the pmp test (from riscv-tests) works.


jerryz123:
The rocket NBDCache currently sets all physically addressed requests to supervisor level. Requests may fail PMP checks in the dcache tlb, even after being checked by the LSU TLB.

This is being fixed in the data cache rework

PR #316, opened 2019-07-05
Branch Predictor Tweaks

bkorpan:
**Type of change**: bug fix | other enhancement

**Impact**: rtl refactoring

**Release Notes**
A collection of tweaks to make BOOM's branch predictor perform at a reasonable level.


src/main/scala/bpu/bpd/gshare/gshare.scala
```diff
@@ -53,6 +53,50 @@ trait HasGShareParameters extends HasBoomCoreParameters
...
+
+  def getUpdated(cfi_idx: UInt, taken: Bool): GShareEntry = {
+    val new_cfi_idx = Wire(UInt(log2Ceil(fetchWidth).W))
+    val new_counter = Wire(UInt(2.W))
+    new_cfi_idx := this.cfi_idx
+
+    when (cfi_idx === this.cfi_idx) {
+      new_counter := updateCounter(taken)
+    } .otherwise {
+      when (isWeak) {
```
abejgonzalez, line 87: I understand that you need to do this so that the new cntr is in the weak state on a switch... however, you might want to add a comment to clarify this?

src/main/scala/bpu/bpd/gshare/gshare.scala
```diff
@@ -63,18 +107,13 @@ trait HasGShareParameters extends HasBoomCoreParameters
...
 {
   val debugIdx = UInt(idxSz.W) // Can recompute index during update (but let's check for errors).
-  val rowdata = UInt((fetchWidth*2).W) // Store to prevent a re-read during an update.
+  val entry = new GShareEntry(fetchWidth) // Store to prevent a re-read during an update.
 
   def isTaken(cfi_idx: UInt) = {
-    val cntr = getCounterValue(cfi_idx)
-    val taken = cntr(1)
+    val counter = entry.counter
+    val taken = counter(1)
```
abejgonzalez, line 120: use `isTaken`?

src/main/scala/bpu/bpd/gshare/gshare.scala
bkorpan: Sure.

src/main/scala/bpu/bpd/gshare/gshare.scala
```diff
@@ -96,72 +135,30 @@ object GShareBrPredictor
...
    with HasGShareParameters
 {
   require (log2Ceil(nSets) == idxSz)
 
-  private def Hash (addr: UInt, hist: UInt) = {
+  private def Hash(addr: UInt, hist: UInt) = {
     // fold history if too big for our table
     val folded_history = Fold (hist, idxSz, historyLength)
-    ((addr >> (log2Ceil(fetchWidth*coreInstBytes).U)) ^ folded_history)(idxSz-1,0)
+    val pc = addr >> log2Ceil(coreInstBytes).U
```
abejgonzalez, line 154: What was the intuition behind changing the hash?

src/main/scala/bpu/bpd/gshare/gshare.scala
bkorpan: Makes sense.

src/main/scala/bpu/bpd/gshare/gshare.scala
bkorpan: It's more common for request PC's to be aligned to fetch-bundle boundaries, so hashing with the 'instruction granularity' bits should result in better utilization.

src/main/scala/bpu/bpd/gshare/gshare.scala
```diff
@@ -53,6 +53,50 @@ trait HasGShareParameters extends HasBoomCoreParameters
   val rowSz = fetchWidth*2
```
abejgonzalez: In the new implementation, you don't need this right?

bkorpan: Right.

src/main/scala/bpu/bpd/gshare/gshare.scala
```diff
@@ -53,6 +53,50 @@ trait HasGShareParameters extends HasBoomCoreParameters
...
 }
 
+/**
+ * Counter table entry.
+ */
+class GShareEntry(val fetchWidth: Int) extends Bundle {
```
abejgonzalez, line 58: What is the impact of having a single counter (with a cfi "tag") with cases of multiple branches in the same fetch packet? Is the assumption that if this happens this is treated just like a hash collision and thus perf. will be bad.

src/main/scala/bpu/btb/btb.scala
```diff
@@ -141,16 +144,20 @@ class BoomBTBResp(implicit p: Parameters) extends BoomBTBBundle
...
+  val taken    = Bool()
 
   // the offset of the PC of the branch
-  val cfi_idx = UInt(log2Ceil(fetchWidth).W)
+  val cfi_idx  = UInt(log2Ceil(fetchWidth).W)
 
   // other branch information
   val bpd_type = BpredType()
   val cfi_type = CfiType()
+
```
abejgonzalez, line 157: Extra space

src/main/scala/bpu/bpd/gshare/gshare.scala
bkorpan: Yeah pretty much. Say you enter a fetch packet with the same history but different position (because an earlier branch mispredicted taken), you'll read and update a different entry based on the hash.

abejgonzalez (approved):
LGTM.

PR #317, opened 2019-07-06
Delayed Dispatch

bkorpan:

**Impact**: rtl refactoring

**Release Notes**
Dispatch into ROB, LSU queues, etc. is delayed until the second stage of rename. This removes any paths which go through decode, the maptable, dispatch hazard-checker, and broadcast into the ROB, which should be good for QoR. This requires the second stage of rename be "rolled back" during pipeline flushes.

The interfaces of the rename structures were also refactored in the process: MicroOps were removed from the interfaces of the maptable and freelist, and their behavior was "simplified". It's not immediately clear that this will result in cleaner code overall, but it should make some planned future features/refactors a bit easier and cleaner.

### DIFF
--- a/src/main/scala/exu/core.scala
+++ b/src/main/scala/exu/core.scala
@@ -464,16 +464,19 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
   //-------------------------------------------------------------
   //-------------------------------------------------------------
 
-  val rocc_shim_busy = if (usingRoCC) !exe_units.rocc_unit.io.rocc.rxq_empty else false.B
-
   // Rename2/Dispatch pipeline logic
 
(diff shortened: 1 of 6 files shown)


jerryz123:
Is this meant to go into beagle. Is this high priority

bkorpan:
> Is this meant to go into beagle. Is this high priority

I wanted to try it out, yeah. We need to see where our QoR is with 4-stage fetch.

bkorpan:
I've booted Fedora with this several times, including a MegaBoomConfig. Should be good to go.

jerryz123:
And this was tested in the beagle tests, so the rocc interface is correct?

bkorpan:
Yeah, it seems to be working as far as we can tell so far.

jerryz123 (approved):

PR #318, opened 2019-07-06
[ren] Handle MV instructions in rename

jerryz123:
**Related issue**: 

**Impact**: new rtl 

**Development Phase**: proposal |  implementation

**Release Notes**

`mv` instructions can be handled by just changing the map-table entry for the destination register, without sending the instruction through to dispatch.

The additional complexity here is reading the intermediate map tables to determine what the new physical destination is (since we don't allocate a new pdst from the freelist). There's a way to do this that is cheaper, but its more complex to implement. 

This conflicts with #317 due to changes to the rename interfaces. I'll rebase this when that is done, since this PR still needs some work.

### DIFF
--- a/src/main/scala/common/consts.scala
+++ b/src/main/scala/common/consts.scala
@@ -278,6 +278,7 @@ trait ScalarOpConstants
   val uopSFENCE    = 107.U(UOPC_SZ.W)
 
   val uopROCC      = 108.U(UOPC_SZ.W)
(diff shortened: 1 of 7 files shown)


bkorpan:
You also need to make sure you don't read a valid stale_pdst until only one lreg is mapped onto that preg. IMO, this is too complicated to be worthwhile right now. There are much lower hanging fruit for performance increases.

Issue #301, opened 2019-06-20
Setup local BuildBot runs


abejgonzalez:
**Type of issue**: feature request

**Impact**: no rtl change

**Development Phase**: request

**Other information**
Currently only uses CircleCI to run regressions on the cores. However, we are bounded by the amount of memory given (https://github.com/riscv-boom/riscv-boom/issues/147 alludes to this). This would be also beneficial to run longer tests.

abejgonzalez:
CI was switched to build remotely so BuildBot is not needed anymore.

PR #321, opened 2019-07-09
Remote builds with CI

abejgonzalez:
**Related issue**: #301 

**Impact**: no rtl change

**Release Notes**
This allows support for remote CI builds on the millennium machines. It builds verilator and the rtl on the servers and sends the build files back to the CI docker instances to run the tests. Additionally, this PR adds back the Large/Mega configs for testing. The supported tests are below (mainly chosen to reduce runtime).

Mega/Large:
- Only bmark tests
- Csmith added


abejgonzalez:
Closed since #322 is now open.

bkorpan (approved):
Should we tweak the number of csmith iterations so that each of the configs finish at around the same time?

jerryz123:
> Should we tweak the number of csmith iterations so that each of the configs finish at around the same time?

Eh. Testing SmallConfig repeatedly isn't that helpful. We probably want the mega configs to finish in ~5 hours or less though

PR #322, opened 2019-07-09
Remote CI Builds

abejgonzalez:
**Related issue**: #301 

**Impact**: no rtl change

**Release Notes**
This allows support for remote CI builds on the millennium machines. It builds verilator and the rtl on the servers and sends the build files back to the CI docker instances to run the tests. Additionally, this PR adds back the Large/Mega configs for testing. The supported tests are below (mainly chosen to reduce runtime).

Mega/Large:
- Only bmark tests
- Csmith added


bkorpan (commented):
Should we tweak the number of csmith iterations so that each of the configs finish at around the same time?

bkorpan (approved):

abejgonzalez:
@bkorpan So I think that the number will vary alot whenever a config changes significantly. So personally don't want to get into the habit of changing it alot. However, I think it is something that can be changed for this PR.

abejgonzalez:
On another note: I am going to mark this PR as "DO NOT MERGE" for a bit. I want to change it so that there is no 'cleanup' step and 'copy' step. This makes it so that there can be dangling files left on the server. Instead, I want to make it so that each step is fully independent of each other.

bkorpan (approved):

.circleci/defaults.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# local variables (aka within the docker container)
+LOCAL_CHECKOUT_DIR=$HOME/project
+LOCAL_RISCV_DIR=$HOME/riscv-tools-install
+LOCAL_CHIPYARD_DIR=$HOME/chipyard
+LOCAL_VERILATOR_DIR=$HOME/verilator
+LOCAL_SIM_DIR=$LOCAL_CHIPYARD_DIR/sims/verisim
+
+# key value store to get the build strings
+declare -A mapping
```
jerryz123, line 35: What is the point of this?

jerryz123 (approved):

PR #328, opened 2019-07-11
Decode QoR

bkorpan:

**Impact**: rtl refactoring

**Release Notes**
Have the fetchbuffer store instruction bits which were expanded in F3, rather than compressed instructions. Decode won't need to re-expand them, saving 50~100ps of delay.


jerryz123:
Sigh this is kind of gross with two `inst` fields in the micro-op bundle. 
Ok for now though.

jerryz123 (approved):

PR #342, opened 2019-07-18
[rocc] Remove debug_inst

bkorpan:

**Impact**: rtl refactoring

**Release Notes**
Use inst instead of debug_inst to get the RoCC instruction. Should prevent debug_inst from being propagated through the fetch buffer post synthesis.


abejgonzalez:
LGTM. Just to be clear the `debug` signals should not be used for logic decisions throughout the core, right?

jerryz123:
This fixes that, originally debug_inst was being used as a logic signal to pass stuff into decode.

jerryz123 (approved):

PR #344, opened 2019-07-21
[core/rob] Get exception PC's from FTQ

bkorpan:

**Impact**: rtl refactoring

**Release Notes**
Have exceptions which were triggered in the frontend fight over the FTQ read port that the branch unit uses. The branch unit has priority, so decode will stall if it contains an excepting instruction and the port is unavailable. This should prevent PC's from propagating through the fetch buffer in a synthesized core.


bkorpan:
Currently broken: hits a kernel panic while booting linux.

bkorpan:
> Currently broken: hits a kernel panic while booting linux.

Fixed it.

jerryz123 (approved):

PR #347, opened 2019-07-22
[ren] Separate INT/FP Rename Instantiations

bkorpan:

**Impact**: rtl refactoring

**Release Notes**
Describe the INT/FP pipelines as separate instantiations of a module, rather than a single instantiation.

### DIFF
--- a/src/main/scala/exu/core.scala
+++ b/src/main/scala/exu/core.scala
@@ -670,7 +670,7 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
     renport.bits := intport.bits
   }
   if (usingFPU) {
-    for ((renport, fpport) <- rename_stage.io.fp_wakeups zip fp_pipeline.io.wakeups) {
+    for ((renport, fpport) <- fp_rename_stage.io.wakeups zip fp_pipeline.io.wakeups) {
        renport <> fpport
     }
   }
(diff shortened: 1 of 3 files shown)


src/main/scala/exu/core.scala
```diff
@@ -438,26 +443,52 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
...
+  }
 
+  // Outputs
   dis_uops := rename_stage.io.ren2_uops
   dis_valids := rename_stage.io.ren2_mask
 
+  /**
+   * TODO This is a bit nasty, but it's currently necessary to
+   * split the INT/FP rename pipelines into separate instantiations.
+   * Won't have to do this anymore with a properly decoupled FP pipeline.
```
jerryz123, line 507: How do you plan on handling int-to-fp/fp-to-int moves without doing this ugly merge step?

bkorpan: If the FP rename pipeline ends up living behind a queue, fp-to-int could be handled the same as it is now. int-to-fp would need to use some sort of structure to match up fp destinations to the uops coming from the int side.

bkorpan:
This boots fedora.

src/main/scala/exu/core.scala
```diff
@@ -104,7 +104,11 @@ class BoomCore(implicit p: Parameters, edge: freechips.rocketchip.tilelink.TLEdg
 
   val decode_units     = for (w <- 0 until decodeWidth) yield { val d = Module(new DecodeUnit); d }
   val dec_brmask_logic = Module(new BranchMaskGenerationLogic(coreWidth))
-  val rename_stage     = Module(new RenameStage(coreWidth, numIntRenameWakeupPorts, numFpWakeupPorts))
+  val rename_stage     = Module(new RenameStage(coreWidth, numIntPhysRegs, numIntRenameWakeupPorts, false))
```
jerryz123, line 107: Nit, but naming this `int_rename_stage` might make more sense,

Although if FP-pipeline becomes more decoupled, this naming is fine,

jerryz123 (approved):

bkorpan:
This needs rebasing for a very simple merge conflict.

PR #353, opened 2019-07-24
[general] Switch BOOM to use TSI by default

abejgonzalez:
**Related issue**: 

**Impact**: rtl refactoring / new rtl

**Release Notes**
This PR changes the default build of BOOM to use TSI/Serial Interface instead of the DTM. This improves the speed of simulations (and CI for that matter) at the cost of making DTM related builds a bit harder to do (must change variables in the `make` build). I added documentation to address this. This also makes it so that https://github.com/ucb-bar/testchipip is a dependency of BOOM. You must address this in the `build.sbt` of the "integration" repo that instantiates BOOM.


jerryz123:
I thought the approach we were going to take was to build the `example` configs in ChipYard, not change the behavior of the `boom` configs.

abejgonzalez:
Thinking about this more... this creates a weird circular dependency where Chipyard depends on BOOM and BOOM depends on Chipyard. So I think the more elegant solution is just to enforce that `testchipip` is included in BOOM.

bkorpan (approved):

abejgonzalez:
Closing since #355 handles this more elegantly

PR #354, opened 2019-07-25
[core] Support breakpoints

### LINKED ISSUE #16: BOOM lacks breakpoint support
BOOM currently lacks breakpoint support.  It would be nice to have it.

Things to note: 

1. The boom branch of rocket-chip disables the `breakpoint` test (so the necessary software to test this feature already exists):

https://github.com/ucb-bar/rocket-chip/blob/boom/src/main/scala/rocketchip/RocketTestSuite.scala#L153

2. The rocket core implements breakpoints and provides a lot of starting pointers for a BOOM implementation:

https://github.com/ucb-bar/rocket-chip/blob/master/src/main/scala/rocket/breakpoint.scala 
https://github.com/ucb-bar/rocket-chip/blob/master/src/main/scala/rocket/rocket.scala#L268

3. One of the challenges is handling a breakpoint on an instruction address in the presence of superscalar fetching.

jerryz123:
**Related issue**: Resolves #16 

**Impact**:  new rtl

**Release Notes**

Breakpoints now supported (I think). We can use Rocket's `DefaultTestSuite` now.


jerryz123:
I should probably double-check this doesn't break buildroot/Fedora.

src/main/scala/ifu/fetch-control-unit.scala
```diff
@@ -28,6 +28,7 @@ import chisel3.util._
 import chisel3.core.{withReset, DontCare}
 import chisel3.experimental.{dontTouch}
 
+import freechips.rocketchip.rocket.{MStatus, BP, BreakpointUnit}
```
bkorpan: The BreakpointUnit isn't used in this file.

bkorpan (commented):
Potential for some QoR impact, but probably alright. I'm not a fan of instantiating a BreakpointUnit in every pipeline slot of the fetch buffer enqueue logic, it seems like this could be done more efficiently, though it might be nasty code.

src/main/scala/ifu/fetch-control-unit.scala
jerryz123: Good catch.

jerryz123:
I defaulted `nBreakpoints` to 0, which should make the BreakPointUnits dummy modules that always output `false`.
When this gets merged, I'll open an issue that mentions updating the frontend breakpointunit.

bkorpan (approved):

abejgonzalez (approved):
LGTM as long as the default behavior is BP = 0 for now.

abejgonzalez:
Lets wait on this PR until the Chipyard PR gets merged in, that will remove the conflict on the BoomTestSuites file

jerryz123:
Fine with me

abejgonzalez:
Why does this still conflict? Did the rebase work correctly?

jerryz123:
> Why does this still conflict? Did the rebase work correctly?

Looks like I messed up. I'll rebase again.

PR #355, opened 2019-07-25
[general] Migrate BOOM Integration to Chipyard

abejgonzalez:
**Related issue**:

**Impact**: rtl refactoring

**Development Phase**:  implementation

**Release Notes**
This PR moves all integration components to Chipyard (https://github.com/ucb-bar/chipyard).  Also moves all config oriented material to the `common` folder. This requires BOOM to transition to using Chipyard over `boom-template`


src/main/scala/common/configs.scala
```diff
@@ -1,254 +1,149 @@
...
-    dcache = Some(DCacheParams(rowBits = site(SystemBusKey).beatBytes*8,
-                               nSets=64, nWays=8, nMSHRs=8, nTLBEntries=32)),
-    icache = Some(ICacheParams(fetchBytes = 4*4, rowBits = site(SystemBusKey).beatBytes*8, nSets=64, nWays=8))
-    )}
-  case SystemBusKey => up(SystemBusKey, site).copy(beatBytes = 16)
-})
 
+// scalastyle:off
+
+// Main configs. SmallBoomConfig and MediumBoomConfig are best-maintained
```
jerryz123, line 47: What are these configs for?

abejgonzalez: These configs are `BOOM` only configs that make it easy for external users to mixin without having to specify each individual mixin that makes it. For ex. in Chipyard if you want a `SmallBoomConfig` with JTAG. You can just do something like

```
class ChipyardSmallBoomConfig {
   WithChipyardIntegrationThatAddsJtag
   SmallBoomConfig
}
```

instead of something like

```
class ChipyardSmallBoomConfig {
   WithChipyardIntegrationThatAddsJtag
   WithRVC
   WithXCores
   ...
   WithRocketBase
}
```

Note: This approach still allows users to specify everything explicitly.

abejgonzalez: We can remove some of these configs but it would bloat the configs file in Chipyard with having to specify all the individual components.

jerryz123: I think leaving this here is fine.

src/main/scala/common/configs.scala
```diff
@@ -1,254 +1,149 @@
 //******************************************************************************
-// Copyright (c) 2015 - 2018, The Regents of the University of California (Regents).
+// Copyright (c) 2015 - 2019, The Regents of the University of California (Regents).
 // All Rights Reserved. See LICENSE and LICENSE.SiFive for license details.
 //------------------------------------------------------------------------------
-// Author: Christopher Celio
+// Author: Christopher Celio, Abraham Gonzalez, Ben Korpan, Jerry Zhao
```
bkorpan, line 5: Not sure if it makes sense to start adding a bunch of names to the author thing. Is listing authors even required by the license?

bkorpan (commented):
csmith seems to have broken because it's getting the name of the sim binary wrong

src/main/scala/common/configs.scala
abejgonzalez: I think the author thing was put by Chris a long time ago... but I don't know what you might define as an author... Is it the first person who created the file? The list of people who modified it? Does moving the file count? Refactoring? I think that the most elegant thing is to just drop the author part across the board and make sure that authors are explictly listed in a README... or GitHub collaborators or something.

bkorpan: I agree.

abejgonzalez: Ok. I think that is for a separate PR. Ill add an issue to track it

bkorpan (approved):

README.md
```diff
@@ -16,158 +16,38 @@ Parameterized |√
...
-````
-For more details (and to download o3-pipeview.py), visit the [gem5 wiki](http://www.m5sim.org/Visualization).
-
-
-## More Info
-
-Check out the BOOM website @ https://boom-core.org!
-
-* **The Design Spec** - https://docs.boom-core.org
-* **Google group:** - https://groups.google.com/forum/#!forum/riscv-boom
```
jerryz123, line 133: Are there links to these in the docs? I didn't see them at a quick glance.

Also leaving the Google group link may be a good idea.

jerryz123 (approved):
LGTM

README.md
abejgonzalez: Line 30 of the new README has a link to the docs + mailing list. Also, these links are not in the docs. I can go ahead and add that.

PR #362, opened 2019-07-30
[exu] intToFp queue size

bkorpan:

**Impact**: rtl refactoring

**Release Notes**
Only need 2 entries in intToFp queue


abejgonzalez:
Do you know why this was being conservative in the first place? Also, can you give a brief reason why its fine to just have 2.

jerryz123:
We can't back pressure the pipeline. The queue has to be large enough for all the non-backpressurable stages to drain into it.

That said, I'm not entirely sure 2 is sufficient? Register-read stage is also non-backpressurable right? so you need the exeUnitLatency + 1 at least?

PR #275, opened 2019-05-29
[lsu] LSU Refactor w. general improvments

jerryz123:
**Type of change**: new feature | other enhancement

**Impact**: new rtl

**Development Phase**:  implementation

**Release Notes**
Changes:
- The DCacheShim is removed, with inflight loads being tracked (and killed on mis-spec) in the DCache itself.
- The HellaCache interface is abandoned in favor of a more decoupled interface.
- A "HellaCacheShim" is implemented in the LSU to handle HellaCache requests from PTW and RoCC. This lets us reuse the LSU TLB, instead of duplicating it in the DCache.
- Boots Linux/Fedora w. SmallBoom, LargeBoom, LargeBoom+L2, DualSmallBooms+L2
- More eager forwarding from store buffer, previous implementation was very conservative
- More conservative memory ordering failures. Try to recover without exception if possible
- PMPs now behave correctly
- Wakeup/retry scheduling is slightly improved
- Cleaner 4-cycle load use implementation
- Reduced L1 miss penalty
- Added `memtracer` utility and print option for debugging loads/stores witnessed by the core
- Forked TLB, which now behaves as nonblocking TLB
- Resolves #251, #170, #165, #122

TODO: (can be completed in this PR or in future ones)
- Update docs/diagrams?
- Add back pipeline state printouts
- Send stores to DCache at PNR
- "Prefetch" stores waiting to execute
- Combine ldq_idx, stq_idx into one field
- Improve atomics performance

PR #365, opened 2019-07-31
Remove the ECC Config

jerryz123:
**Related issue**: 

**Type of change**: bug fix

**Impact**: no rtl change 

**Release Notes**


abejgonzalez:
Why do we want to remove the ECC config? Is this associated with #275 i.e. new LSU doesn't support it yet.

jerryz123:
New DCache does not support it. Support for ECC in old DCache was questionable. 
Also we would probably only be able to support correctable ECC anyways.

abejgonzalez (approved):

PR #373, opened 2019-08-23
[ren] Check register type matches for bypassed allocations

jerryz123:
**Related issue**: 

**Type of change**: bug fix


bkorpan (commented):
Shouldn't need to do this afaik, has this triggered a bug?

jerryz123:
This triggered a bug. I don't see where the bypass bsy bit is gated by register type?
There was a fcvt instruction which was having prs2_bsy set, I assume because if the register type were RT_X it would use the integer rename busy output

bkorpan:
I think I see the problem, it should do the busy bit selection differently.

jerryz123:
If you have a more correct solution I'm happy to use that one.

bkorpan:
Try #374

PR #376, opened 2019-08-25
[ren] Fix RT_PAS passthrough

jerryz123:
**Type of change**: bug fix


jerryz123:
Confirmed this works. Merging due to to this blocking other things

PR #378, opened 2019-08-26
banked dcache

bkorpan:
verilator/vcs sim compilation tends to hang on 128-bit bus configs for some reason

### DIFF
--- a/src/main/scala/common/parameters.scala
+++ b/src/main/scala/common/parameters.scala
@@ -67,6 +67,7 @@ case class BoomCoreParams(
   nPerfCounters: Int = 0,
   numRXQEntries: Int = 4,
   numRCQEntries: Int = 8,
+  numDCacheBanks: Int = 1,
   /* more stuff */
 
   useFetchMonitor: Boolean = true,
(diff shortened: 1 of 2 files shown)


jerryz123:
I've noticed that too. MegaBoom dual-mem takes 2 hours for VCS compile.

@bkorpan How did you find that VecInit.tabulate was causing hangs?

bkorpan:
> I've noticed that too. MegaBoom dual-mem takes 2 hours for VCS compile.

I fought with it for days thinking it a bug. Does the VCS step eventually finish? The part it was hanging on normally takes a couple seconds.

jerryz123:
> > I've noticed that too. MegaBoom dual-mem takes 2 hours for VCS compile.
> 
> I fought with it for days thinking it a bug. Does the VCS step eventually finish? The part it was hanging on normally takes a couple seconds.

It does eventually finish. I have no idea exactly what might be blowing up the compile time

src/main/scala/common/parameters.scala
```diff
@@ -67,6 +67,7 @@ case class BoomCoreParams(
   nPerfCounters: Int = 0,
   numRXQEntries: Int = 4,
   numRCQEntries: Int = 8,
+  numDCacheBanks: Int = 1,
```
jerryz123, line 70: What is the behavior when `numDCacheBanks=1`, `memWidth=2`. Does the load to the second pipe slot always get nack'd?

bkorpan: `require (nBanks >= memWidth)`

PR #379, opened 2019-08-26
Move system/subsystem/configs to chipyard

jerryz123:
Merge this with https://github.com/ucb-bar/chipyard/pull/211


src/main/scala/common/config-mixins.scala
```diff
@@ -112,16 +90,6 @@ class WithTrace extends Config((site, here, up) => {
   case BoomTilesKey => up(BoomTilesKey, site) map { b => b.copy(trace = true) }
 })
 
-/**
```
abejgonzalez, line 115: Maybe make this a "withoutRVC"? I think that will make it cleaner for the person doing the mixins.

jerryz123: I can add the mixin, but given that no one is building a config that does not support RVC, this mixin may never get used.

abejgonzalez: Ok. I don't know if there is ever a case to create the core without RVGC... I think all app cores should probably have GC support.

abejgonzalez (approved):
LGTM. I'm surprised that you need to double the timeout time.

.circleci/config.yml
```diff
@@ -170,7 +170,7 @@ jobs:
             - run:
                 name: Building MegaBoomConfig using Verilator
                 command: .circleci/do-rtl-build.sh megaboom
-                no_output_timeout: 240m
+                no_output_timeout: 480m
```
abejgonzalez: Since this is theoretically fixed by building in parallel... Can we just revert this to what it was before... and/or make it lower now that Verilator builds quickly?

jerryz123: I'll do that once the current CI passes, I'll see what a reasonable time limit is

jerryz123: Done. All builds finish in <1 hour, but I set it to 2h max in case the build server is overloaded

abejgonzalez (approved):
LGTM. It's great that Verilator sim now compiles quickly! 150/100% amazing!

PR #383, opened 2019-09-16
Ci order fix

jerryz123:
Longer jobs should be scheduled first


jerryz123:
Turns out the YML is unordered, this doesn't work

PR #384, opened 2019-09-16
[tile] Flatten BOOM Tile organization

jerryz123:
Flatten the RTL for the BoomTile out of separate traits into just the BoomTile module.


src/main/scala/common/tile.scala
```diff
@@ -179,38 +176,48 @@ class BoomTile(
...
+
+  // ROCC
+  val roccs = p(BuildRoCC).map(_(p))
+  roccs.map(_.atlNode).foreach { atl => tlMasterXbar.node :=* atl }
+  roccs.map(_.tlNode).foreach { tl => tlOtherMastersNode :=* tl }
+
 }
 
 /**
  * BOOM tile implicit
```
abejgonzalez: We should change this naming that I did when I didn't know `Imp` == `Implementation`.

abejgonzalez (approved):
LGTM

PR #387, opened 2019-09-17
Disallow UnifiedMemInt issue queues

jerryz123:
**Related issue**: 

**Type of change**: deprecate feature

**Impact**: no rtl change

Disallow Unified MEM+INT issue queues.

This parameter makes it difficult to do optimizations on the MEM issue queue for >1 translation width.


abejgonzalez (approved):
LGTM

PR #391, opened 2019-09-19
[core] Always enable fast wakeups to rename stage

jerryz123:
**Related issue**: 
Always enable fast wakeups to rename. One less configuration option to maintain.

**Impact**: no rtl change


abejgonzalez (approved):
LGTM

PR #393, opened 2019-09-19
[core] Refactor CfiTypes constants

jerryz123:
This more closely matches the organization of constants in the repo.

### DIFF
--- a/src/main/scala/bpu/btb/btb.scala
+++ b/src/main/scala/bpu/btb/btb.scala
@@ -126,7 +126,7 @@ class BoomBTBResp(implicit p: Parameters) extends BoomBTBBundle
   //   shifting off the lowest log(inst_bytes) bits off).
   val cfi_idx   = UInt(log2Ceil(fetchWidth).W) // where is cfi we are predicting?
   val bpd_type  = BpredType() // which predictor should we use?
-  val cfi_type  = CfiType()  // what type of instruction is this?
+  val cfi_type  = UInt(CFI_SZ.W)  // what type of instruction is this?
   val fetch_pc  = UInt(vaddrBits.W) // the PC we're predicting on (start of the fetch packet).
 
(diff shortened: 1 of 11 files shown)


src/main/scala/common/consts.scala
```diff
@@ -83,6 +84,13 @@ trait ScalarOpConstants
   //************************************
   // Control Signals
 
+  // CFI types
+  val CFI_SZ   = 3
```
abejgonzalez, line 88: Can we add comments on what X means

src/main/scala/bpu/btb/btb.scala
```diff
@@ -126,7 +126,7 @@ class BoomBTBResp(implicit p: Parameters) extends BoomBTBBundle
   //   shifting off the lowest log(inst_bytes) bits off).
   val cfi_idx   = UInt(log2Ceil(fetchWidth).W) // where is cfi we are predicting?
   val bpd_type  = BpredType() // which predictor should we use?
```
abejgonzalez, line 128: If we do this PR (which I do like), we should change the `BpredType` to a similar format.

jerryz123: I consider the branch prediction code to be legacy at this point. You can see in the `ifu2` branch this is all deleted. I don't want to change this here, because it will introduce merge and rebase conflicts for that branch

abejgonzalez: Ok. If that is the plan then that is fine.

src/main/scala/common/consts.scala
```diff
@@ -68,6 +68,7 @@ trait IQType
   val IQT_MFP = 6.U(IQT_SZ.W)
 }
 
+
```
abejgonzalez, line 71: Extra space? Looks like the rest of the file only has single spaces.

abejgonzalez (approved):
Other than the space. LGTM.

PR #394, opened 2019-09-19
[core] Remove unsupported parameters

jerryz123:
FetchLatency and RenameLatency are fixed, and we always register branch resolutions in practice. 

@bkorpan @abejgonzalez can we also remove the unordered-issue queue option? Or the custom-rf-model option?


abejgonzalez:
I think we can remove the `custom-rf` since we will have to do something more unique in the future to deal with physical constraints (more ports, etc). We can keep it if we don't care about the "extra" parameter... I don't think it matters really. As for the IQ priority... I think we can keep both. Even though we use the age-ordered IQ, maybe we might want lower power IQ's in the future (where we don't necessarily R/W every slot every cycle)

abejgonzalez (approved):
LGTM.

PR #395, opened 2019-09-19
[core] Clean up how load-misses kill misspeculated load-wakeups

jerryz123:
This fixes the issue of the `uop.iw_p1_poisoned` bits propagating far into the pipeline, when really they should only propagate into issue.


abejgonzalez (approved):
LGTM

bkorpan (approved):

PR #396, opened 2019-09-20
[core] uop.ctrl.rf_wen is just an alias for uop.dst_rtype

jerryz123:
Just check dst_rtype instead of keeping around a separate signal.


abejgonzalez (approved):
Seems reasonable. I think we should commend on all the control signals in `consts.scala` though.

jerryz123:
I want to split the control signals out of the micro-op bundle, because they should not be passed around the pipeline. This is the first step towards that.

PR #397, opened 2019-09-22
[iss] Split the issue unit generation

jerryz123:
KISS


abejgonzalez (approved):
LGTM. On another note, maybe we should rename "Issue Unit" to "Issue Queue" to align with literature more (and our docs).

PR #401, opened 2019-09-25
[rocc] Fix RoCC shim not getting proper flushes

jerryz123:
Bad naming at fault here.


jerryz123:
Merging so we can get a stable version for Chipyard

PR #405, opened 2019-10-13
[mshrs] Fix tag update hazard for refills

jerryz123:
Weird this only acted up with Howie's prefetcher.

**Type of change**: bug fix


abejgonzalez (approved):
LGTM

PR #413, opened 2019-11-18
[exu] Size the badvaddr reg properly to make sign-extending work.

jerryz123:
**Related issue**: 
Fixes Linux 5.3 with RVC kernels on FireSim. 

**Type of change**: bug fix


abejgonzalez (approved):
LGTM. Looks like it boots Fedora, and runs Coremark on Buildroot

PR #418, opened 2019-12-03
[lsu] AMOs should generate pf/ae_st. Do this by putting LR in the LDQ

jerryz123:
LRs now no longer sit in the STQ. We mark them `is_unique` so they act as barriers.
Now all STQ uops generate store exceptions, while all LDQ uops generate load exceptions.


abejgonzalez:
Fixes #417

abejgonzalez (approved):
LGTM. Boots Linux/Coremark/Fedora

PR #434, opened 2020-01-23
Connect local clock and reset to each element in trace port

abejgonzalez:
**Related issue**: 

**Type of change**: bug fix

**Impact**: new rtl

**Release Notes**
Copy of #433 that should pass CI. The issue was that the CI would run on the #433 PR and would have an older version of Chipyard that is incompatible with the current version of CI.


jerryz123:
I just merged the other one instead

PR #437, opened 2020-02-06
[exu] Suppress illegal instructions from dispatching into the LSU or the RoCC queue

jerryz123:
**Type of change**: bug fix

**Impact**:  rtl refactoring


sagark (approved):
this fixed my problem, tested with rocc accel in firesim simulation, booting linux

PR #446, opened 2020-03-02
Check for mask overlap before flagging st->ld ordering failures

bkorpan:


jerryz123:
Fancy. 
Did you observe this actually affecting performance anywhere?

jerryz123 (approved):

bkorpan:
> Fancy.
> Did you observe this actually affecting performance anywhere?

On my fork of boom. Separate words in the same dword were frequently being flagged.

PR #460, opened 2020-05-05
Fixed false btb clear

pentin-as:
When there was stall (!req_valid), then is_valid remained in the last state.
However, rtag continued to change.
As a result there could be false hits_oh and clear_valid.

**Type of change**: bug fix

**Impact**: new rtl

**Release Notes**


jerryz123:
A new BTB design will be merged in soon, where this bug is not present.

abejgonzalez:
The new BOOM updates are now merged in. This should be good to close now. Feel free to re-open if this is still an issue.

PR #468, opened 2020-06-01
[lsu] Fix for RoCC accels which write L1

jerryz123:
**Related issue**: 
Fix for RoCC accelerators which write L1


jerryz123:
This will go into 3.1, but I'll leave the PR open so we can see the CI run on it.

PR #476, opened 2020-07-18
Ultrawide Core Fixes

bkorpan:
Add configs with 256-bit fetch and >2 memory pipes.


jerryz123:
Does this actually work? Its pretty cool if it does.

I suspect there may be some bug in the global history maintenance logic that reduces bpred performance, I assume its 2x 128b banks?
Also, I'm a bit concerned about moving the store pipe to the last pipe in the cache, that was extremely finicky to get right, but I can't think of a case of the top of my head where it might break something.

bkorpan:
> 
> 
> Does this actually work? Its pretty cool if it does.
> 
> I suspect there may be some bug in the global history maintenance logic that reduces bpred performance, I assume its 2x 128b banks?
> Also, I'm a bit concerned about moving the store pipe to the last pipe in the cache, that was extremely finicky to get right, but I can't think of a case of the top of my head where it might break something.

I was able to run Dhrystone before doing the store pipe thing - got a score of 11500 Dhrystones/MHz. The store pipe thing doesn't work yet.

PR #484, opened 2020-08-20
[mshr] MSHRS should not wait on prober when not to same set

jerryz123:


src/main/scala/lsu/dcache.scala
```diff
@@ -777,7 +782,7 @@ class BoomNonBlockingDCacheModule(outer: BoomNonBlockingDCache) extends LazyModu
   metaWriteArb.io.in(1) <> prober.io.meta_write
   prober.io.mshr_rdy    := mshrs.io.probe_rdy
   prober.io.wb_rdy      := (prober.io.meta_write.bits.idx =/= wb.io.idx.bits) || !wb.io.idx.valid
-  mshrs.io.prober_idle  := prober.io.req.ready && !lrsc_valid
```
zhemao, line 780: Since you're removing `!lrsc_valid`, how do you handle the condition in which `prober.io.state.valid` is low, `lrsc_valid` isn't?

zhemao (approved):
LGTM

PR #543, opened 2021-05-01
lsu,mshr: fix not fully initialized issue for atomics

sashimi-yzh:
**Related issue**: 

**Type of change**: bug fix

**Impact**: no rtl change

**Release Notes**

In chisel3, we should use `0.U.asTypeOf(...)` to initialize the signal.


jerryz123:
Thanks. Usually we integrate into systems with atomic-supporting managers, so this case was never tested before.

jerryz123 (approved):

PR #588, opened 2022-01-11
Added width to UInt in FetchBundle

pentin-as:


pentin-as:
In my environment, the project was not built without this correction

PR #601, opened 2022-05-08
Fix decode from rocket-chip new opcode

ZenithalHourlyRate:
**Related issue**: https://github.com/chipsalliance/rocket-chip/pull/2972

**Type of change**: bug fix

Since https://github.com/riscv/riscv-opcodes/pull/106 the API of riscv-opcodes has changed, especially on `SLLI_RV32` and RoCC custom instructions. Rocket Chip has bumped to new API, BOOM here also needs to bump them.

**Impact**: no rtl change

**Release Notes**

Fixed rocketchip Instructions API since https://github.com/chipsalliance/rocket-chip/pull/2972


jerryz123 (approved):
thanks

PR #621, opened 2022-10-30
make some changes to improve the clarity of expression

Eddiewdy:
**Related issue**: 

**Impact**: unknown

**Release Notes**

when reading the file config-mixins.scala, it got really confusing to understand the relation between cache fetchByte and core fetchWidth. So I think maybe it's better to point out the relation between these two params.


Eddiewdy:
And why it uses 2 * 2 when fetchWidth is 4? Why use multiply instead of just 4? Also can find this issue where fetchWidth is 8.

Issue #628, opened 2023-02-15
Bug report: Static rounding mode into account in FDivSqrtUnit


flaviens:
Hi there!

I discovered a bug in BOOM.
In the FPU, the following combinations of (instruction, static rounding mode, dynamic rounding mode) lead, depending on the operands, to erroneous results due to bad rounding.

```
fdiv.d
    static: 0, fcsr: 1 
    static: 0, fcsr: 2 
    static: 0, fcsr: 3 
    static: 1, fcsr: 0 
    static: 1, fcsr: 2 
    static: 1, fcsr: 3 
    static: 2, fcsr: 0 
    static: 2, fcsr: 1 
    static: 2, fcsr: 3 
    static: 3, fcsr: 1 
    static: 3, fcsr: 2 
    static: 3, fcsr: 4 
    static: 4, fcsr: 2 
    static: 4, fcsr: 3 
fsqrt.d
    static: 0, fcsr: 1
    static: 1, fcsr: 0
    static: 1, fcsr: 3
    static: 1, fcsr: 4
    static: 2, fcsr: 0
    static: 2, fcsr: 3
    static: 2, fcsr: 4
    static: 3, fcsr: 1
    static: 3, fcsr: 2
    static: 3, fcsr: 4
    static: 4, fcsr: 1
    static: 4, fcsr: 2
    static: 4, fcsr: 3
```

We can see that the mismatch happens only when static != fcsr.

**RTL implications**
Minor RTL modification.

**What is the use case for changing the behavior?**
Correct rounding results in all cases.

**Notes**
This bug very probably encompasses an observation made in #564

PR #629, opened 2023-02-15
Bugfix: take static rounding mode into account in FDivSqrtUnit

flaviens:
Hi there!

**Related issue**: #628 

**Type of change**: bug fix

**Impact**: rtl refactoring

**Release Notes**
Take static rounding mode into account in FDivSqrtUnit

**Testing**
I tested the fix with many inputs and all rounding errors disappear.


jerryz123 (approved):
Thanks for the fix!

PR #643, opened 2023-06-05
[lsu] Fix logic to prevent forwarding from AMO instructions

plutsyk:
fix control logic to prevent forwarding from AMOs

**Type of change**: bug fix

**Impact**: rtl refactoring

**Release Notes**

When executing load instructions, the store queue is searched for forwarding candidates. This search should account for possible address overlapping AMO instructions only to terminate memory accesses, not to use their data for forwarding. To implement this behavior, it suffices to change the store queue search implementation at the second **_when_** statement s.t. for store queue entries containing AMOs its second condition triggers instead of the first.

**DEVS ONLY: FORKED PR**
Developers should use https://github.com/jklukas/git-push-fork-to-upstream-branch (or similar mechanism) to trigger CI for this PR before merging.


jerryz123 (approved):
Thanks!

PR #644, opened 2023-06-21
Put terminology in alphabetical order

pepper-lees:
**Related issue**: 

**Type of change**: other enhancement (documentation)

**Impact**: no rtl change 

**Development Phase**: N/A

Since there's more than a small handful of terms in the terminolgoy, I think it'd be helpful to put them in alphabetical order. I know that it's sometimes convenient to have relevant terms next to each other (e.g. front-and and back-end), but since there's more than 20 items in the list, it starts becoming more easily searchable if they're in order.

There's a joke in here somewhere about putting things in order for the out-of-order machine, but I can't think of a punchline.

### DIFF
--- a/docs/sections/terminology.rst
+++ b/docs/sections/terminology.rst
@@ -6,6 +6,42 @@ out-of-order terminology.
 
 .. glossary::
 
+    Back-end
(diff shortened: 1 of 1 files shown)


jerryz123 (approved):
Thanks!

PR #656, opened 2023-09-16
regfile: switch from Mem() to SyncReadMem()

oharboe:
**Related issue**: 

**Type of change**: bug fix | new feature | other enhancement

other enhancement

**Impact**: no rtl change | rtl refactoring | new rtl | unknown

rtl refactoring

**Development Phase**: proposal |  implementation

implementation

**Release Notes**

Switch RegisterFile from Mem() to SyncReadMem()(from asynchronous read to synchronous read memory)

### DIFF
--- a/src/main/scala/exu/register-read/regfile.scala
+++ b/src/main/scala/exu/register-read/regfile.scala
@@ -114,7 +114,7 @@ class RegisterFileSynthesizable(
 {
   // --------------------------------------------------------------
 
-  val regfile = Mem(numRegisters, UInt(registerWidth.W))
+  val regfile = SyncReadMem(numRegisters, UInt(registerWidth.W))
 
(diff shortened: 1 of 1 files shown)


oharboe:
Passed this test in sims/verilator:

```
make CONFIG=MegaBoomConfig run-asm-tests
```

What other tests can I run?

oharboe:
No further followup on this. The problem is that all SyncReadMem()s are converted to memories automatically and for the registerfile something else than a normal SRAM is needed since the registerfile has so many ports. Using Mem() instead of SyncReadMem() avoids the registerfile being swept up by Chipyard in macros that should be converted to use SRAMs. This is a little bit of a fudge, but using SyncReadMem() for registerfile needs to come with some way to avoid it being built of normal SRAMs.

PR #706, opened 2024-10-01
fix: fix an ordering issue that blocks pnr

tianrui-wei:
This commit fixes an ordering issue such that under these scenarios, PNR will be broken.

1. an older load misses
2. a younger load gets issued after the cacheline comes back but before the older load gets replayed
3. immediately after the younger load hits, the line gets evicted due to snooping.

This breaks ld-ld ordering for the same address, and although the old line conceives no memory ordering violation could be made, it still needs to be replayed, thus blocking PnR.

**Related issue**: 

**Type of change**: bug fix

**Impact**: rtl refactoring

**Release Notes**

### DIFF
--- a/src/main/scala/v4/lsu/lsu.scala
+++ b/src/main/scala/v4/lsu/lsu.scala
@@ -1240,8 +1240,10 @@ class LSU(implicit p: Parameters, edge: TLEdgeOut) extends BoomModule()(p)
           when ((l_executed || l_succeeded) &&
                 !s1_executing_loads(i) && // If the load is proceeding in parallel we don't need to kill it
                 l_observed) {        // Its only a ordering failure if the cache line was observed between the younger load and us
-            ldq_order_fail(i) := true.B
-            failed_load := true.B
+            //ldq_order_fail(i) := true.B
+            // failed_load := true.B
+
+	    // this case is no longer possible where an older search gets replayed and found that a younger search has already been released. Note that this is stronger than rvwmo.
             assert(false.B)
           }
         } .elsewhen (lcam_ldq_idx(w) =/= i.U) {
--- a/src/main/scala/v4/lsu/lsu.scala
+++ b/src/main/scala/v4/lsu/lsu.scala
@@ -320,6 +320,11 @@ class LSU(implicit p: Parameters, edge: TLEdgeOut) extends BoomModule()(p)
 
   def widthMap[T <: Data](f: Int => T) = VecInit((0 until lsuWidth).map(f))
 
+  // default connection from flip flop to next state
+
+  val ldq_will_succeed        = WireDefault(ldq_succeeded)
+  ldq_succeeded := ldq_will_succeed
+
 
   //-------------------------------------------------------------
   //-------------------------------------------------------------
@@ -393,7 +398,7 @@ class LSU(implicit p: Parameters, edge: TLEdgeOut) extends BoomModule()(p)
       ldq_uop            (ldq_idx)       := UpdateBrMask(io.core.brupdate, dis_uops(w).bits)
       ldq_addr           (ldq_idx).valid := false.B
       ldq_executed       (ldq_idx)       := false.B
-      ldq_succeeded      (ldq_idx)       := false.B
+      ldq_will_succeed   (ldq_idx)       := false.B
       ldq_order_fail     (ldq_idx)       := false.B
       ldq_observed       (ldq_idx)       := false.B
       ldq_forward_std_val(ldq_idx)       := false.B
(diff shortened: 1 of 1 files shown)


src/main/scala/v4/lsu/lsu.scala
```diff
@@ -1236,11 +1236,12 @@ class LSU(implicit p: Parameters, edge: TLEdgeOut) extends BoomModule()(p)
                 l_observed) {        // Its only a ordering failure if the cache line was observed between the younger load and us
             ldq_order_fail(i) := true.B
             failed_load := true.B
+            assert(false.B)
```
jerryz123, line 1247: Why is this here?

src/main/scala/v4/lsu/lsu.scala
```diff
@@ -1236,11 +1236,12 @@ class LSU(implicit p: Parameters, edge: TLEdgeOut) extends BoomModule()(p)
...
           }
         } .elsewhen (lcam_ldq_idx(w) =/= i.U) {
           // The load is older, and it wasn't executed
           // we need to kill ourselves, and prevent forwarding
-          when (!(l_executed || l_succeeded)) {
+          when (!(l_executed && l_succeeded)) {
```
jerryz123: While I believe this might fix the bug, what is the implication on back-to-back loads that hit in the cache? Will the second load be killed here?

src/main/scala/v4/lsu/lsu.scala
tianrui-wei: this patch prevents the case that when an older load gets issued, it will not observe a younger load that has been observed. Hence the assertion.

jerryz123: Can you add a comment indicating that this case is impossible (and comment out the assignments to ldq_order_fail and failed_load. This will remove the unnecessary logic to set those wires.

src/main/scala/v4/lsu/lsu.scala
tianrui-wei: That's a good point, I'll think about how to fix it. I think this will prevent back to back loads from working sadly

tianrui-wei: I added a fix. Could you take a look at it?

src/main/scala/v4/lsu/lsu.scala
tianrui-wei: Done.

jerryz123:
Can you make a PR into chipyard updating boom to this version? That will trigger all the Chipyard tests. Also, you can push your branch directly into this repo now

tianrui-wei:
Chipyard PR at https://github.com/ucb-bar/chipyard/pull/2071

src/main/scala/v4/lsu/lsu.scala
```diff
@@ -1270,7 +1281,9 @@ class LSU(implicit p: Parameters, edge: TLEdgeOut) extends BoomModule()(p)
           io.dmem.s1_kill(w) := true.B
         }
         kill_forward(w) := true.B
-      }
+	    }
```
jerryz123: Can you fix this brace?

tianrui-wei: Fixed!

jerryz123 (approved):

jerryz123:
FWIW, creating a combinational path from the end of the DCache response stage, into the end of the CAM stage, makes me nervous. It might be ok, but :shrug:

tianrui-wei:
That is a good point. If this ever comes up as a problem, we can delay the kill signal by one cycle and then `and` the signals for kill. I'll merge this in for now and test if this indeed fixes the fastpnr. Do you want me to leave the cy PR open?

src/main/scala/v4/lsu/lsu.scala
Heresyrac: > While I believe this might fix the bug, what is the implication on back-to-back loads that hit in the cache? Will the second load be killed here?

This implementation has significant  performance issues. I found that testcase where load-store-load operations at the same address are repeated, this modification results in a 26% performance loss. The reason is as follows:

1. Older loads successfully match with previous stores and retrieve data from the STQ. At this point, the executed flag is not set to 1, meaning even though the data has been successfully retrieved, the condition executed && succeed is not met.
2. Younger loads check the LDQ and find an older load at the same address with executed=0, triggering a kill_forward.
The younger load repeatedly goes through wakeup-kill_forward until the older load commits and leaves the LDQ, at which point kill_forward is no longer triggered.
3. The younger load completes, resulting in significant additional latency.
I have already raised a new issue to explain this problem.

tianrui-wei: Hi @Heresyrac, thank you for raising this issue. I'll look into this today.

Issue #690, opened 2024-05-08
Dual boom core simulation hang


crsullivan13:
**Type of issue**: bug report

**Impact**: unknown

**Development Phase**: request

**Other information**

https://github.com/riscv-boom/riscv-boom/issues/659 seems related.c
https://groups.google.com/g/chipyard/c/rOhedyr7O-4 opened discussion on the Chipyard forms.
Simulating on FireSim version 1.17.1. The configuration I am simulating is the following:
```
class FsimDualBoom2BankConfig extends Config(
  new freechips.rocketchip.subsystem.WithNBanks(2) ++
  new freechips.rocketchip.subsystem.WithInclusiveCache(nWays=16, capacityKB=1024) ++
  new freechips.rocketchip.subsystem.WithExtMemSize((1 << 30) * 4L) ++
  new WithDefaultFireSimBridges ++
  new WithDefaultMemModel ++
  new WithFireSimConfigTweaks ++
  new chipyard.DualLargeBoomConfig)
  
 class DualLargeBoomConfig extends Config(
  new chipyard.config.WithNPerfCounters(6) ++
  new boom.common.WithNLargeBooms(2) ++                          // 2 boom cores
  new chipyard.config.WithSystemBusWidth(128) ++
  new chipyard.config.AbstractConfig)
```
This issue has fired the pipeline hang assertion in Boom.

To reproduce, simulate the above design on FireSim. Run either two Disparity workloads from SD-VBS with the image input size of CIF or run two simple sequential write tasks that are LLC fitting (https://github.com/CSL-KU/IsolBench/blob/master/bench/bandwidth.c).

I should note that other pairings of SD-VBS workloads (which have lower LLC bandwidth than Disparity) do not experience the hang.

**What is the current behavior?**
Currently, when either of these workload pairs are run, the simulation hangs indefinitely with the pipeline hang assertion firing if synthesizable asserts are enabled.

**What is the expected behavior?**
These seem like normal workloads to want to simulate. The expected behavior is that the simulation would not hang.

**Please tell us about your environment:**

Boom commit hash 1a153d4.
Chipyard commit hash a88d69b.
os: Linux buildroot 6.2.0-00004-ge3e5d275a34f-dirty #25 SMP Tue May  7 22:04:17 CDT 2024 riscv64 GNU/Linux

**What is the use case for changing the behavior?**
These seem like perfectly reasonable workloads to run in a simulation environment without any issue.

jerryz123:
[...]

PR #723, opened 2025-02-27
Fix d$ deadlock with rocket inclusive cache

crsullivan13:
**Related issue**: https://github.com/riscv-boom/riscv-boom/issues/690

**Type of change**: Bug fix

**Impact**: new rtl

**Release Notes**

This fixes a scenario where the d$ MSHR could deadlock with the rocket inclusive cache. Issue was produced in FireSim. Fix has been tested on small, medium and large quad-core boom configurations for v3 BOOM. Haven't tested on v4, but it should be the same and can port if needed.

Note that this PR also includes fixing the "fromSource" field of probe responses. The fix was tested with this, but I am not certain it needs to be there for the fix to work.

Part of this change could open an interference channel in the cores as probes are now blocked when the d$ write-back unit is busy. Have yet to explore this.


ed-schwarz:
Does this solve the problem from #690 ?

crsullivan13:
@ed-schwarz yes it does, I made this fix because I found issue #690. We have been using BOOMv3 modified with this patch for a while now without issue and recently published a paper that required this fix (needed to run multiple write heavy workloads).

ed-schwarz:
How did you find where the panic happens? I think that I am having a similar issue, but unfortunatly your fixes didn't solve. I am struggling a bit with the debbuging to find exactly what happens. If you can share which setup you used and how you captured the bug it would be very helpful.

crsullivan13:
@ed-schwarz I assume you are using FireSim? Is there an assertion firing? This issue showed up via the pipeline hang assertion [https://github.com/riscv-boom/riscv-boom/blob/af3023c5ef22d6a19e2f4608b29f61c37419c115/src/main/scala/v3/exu/core.scala#L1330](url).

The last comment in issue #690 shows the exact config setup that I was using. We work using FireSim on a local VCU118 FPGA. To actually find the root of the problem I was dumping waveforms from Verilator metasimulation. I am surprised that you are having the same issue as this patch has fixed it completely for us.

I am happy to help you figure out what is going on. My email is connor.sullivan13@ku.edu. Feel free to reach out there if you want to talk more so we don't fill up this pr thread.

PR #774, opened 2026-06-15
fix: exec & dcache asserts to be more permissive

jimfangx:
RTL modifications by @Ferran-Hermida-Rivera, upstreaming from internal dev work for upcoming chipyard release.

Modifications boot linux & pass spec 2017

**Related issue**: 

**Type of change**: bug fix

**Impact**: rtl change

**Release Notes**

### DIFF
--- a/src/main/scala/v3/lsu/dcache.scala
+++ b/src/main/scala/v3/lsu/dcache.scala
@@ -700,7 +700,7 @@ class BoomNonBlockingDCacheModule(outer: BoomNonBlockingDCache) extends LazyModu
       }
     }
   }
-  assert(debug_sc_fail_cnt < 100.U, "L1DCache failed too many SCs in a row")
+  assert(debug_sc_fail_cnt < 100000.U, "L1DCache failed too many SCs in a row")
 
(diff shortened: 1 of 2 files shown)


Ferran-Hermida-Rivera:
`dfma latency` fix originally written by @iansseijelly.

src/main/scala/v3/lsu/dcache.scala
```diff
@@ -700,7 +700,7 @@ class BoomNonBlockingDCacheModule(outer: BoomNonBlockingDCache) extends LazyModu
       }
     }
   }
-  assert(debug_sc_fail_cnt < 100.U, "L1DCache failed too many SCs in a row")
+  assert(debug_sc_fail_cnt < 100000.U, "L1DCache failed too many SCs in a row")
```
Ferran-Hermida-Rivera, line 703: `debug_sc_fail_cnt` is an 8-bit counter. It needs to be increased for the assertion to be useful or the limit decreased (otherwise it's effectively deactivated).
Some more debugging needs to be done to understand why it needs more than a 100 retries.

Ferran-Hermida-Rivera: Done.

jerryz123 (approved):

PR #789, opened 2026-07-22
[lsu] Gates the memory excpetion signals with tlb miss so they except…

Ferran-Hermida-Rivera:
…ions are not triggered on invalid/unproduced tlb accesses

**Related issue**: 
Very specific cases in the DTLB might generate an exception signal when the original access was a miss. The LSU doesn't ignore the exception and ends up rising the exception.

**Type of change**: bug fix

**Impact**: new rtl

**Release Notes**


tianrui-wei (approved):
Lgtm

